PR #8, opened 2013-08-24
Cannot rely on HTTP_HOST 

TomHAnderson:
https://bugsnag.com/errors/5217b6becf4cfd7373c945d8/events/newest

The above bug from my account shows

PHP Notice: Undefined index: HTTP_HOST

```
/var/lib/openshift/5217b6484382ecd56c001d7d/app-root/runtime/repo/vendor/bugsnag/bugsnag/src/Bugsnag/Request.php:70 • [main]
```

### DIFF
--- a/src/Bugsnag/Request.php
+++ b/src/Bugsnag/Request.php
@@ -64,10 +64,17 @@ class Request {
         }
     }
 
-    private static function getCurrentUrl() {
+    private static function getCurrentUrl() 
+    {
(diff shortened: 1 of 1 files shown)


drekinov:
Hi, that is very old issue but still valid.
https://bugsnag.com/aikidesk-ltd/ad-www/errors/577ba99205b6c8c6cf49cce3

we are using php 7.0.x (latest) and we get logged exceptions from random IPs around the world (probably HTTP 1.0 bots)
http://php.net/manual/en/reserved.variables.server.php 

> Contents of the Host: header from the current request, if there is one.

We are using 2.x but as i see even 3.x use same HTTP_HOST without checking for availability.

GrahamCampbell:
Does https://github.com/bugsnag/bugsnag-php/pull/302 fix this for you please?

PR #75, opened 2014-12-31
Added the 'setHostname' method to allow an app to set a custom hostname ...

erictompkins:
...for cases where it might not be automatically detected correctly.


ConradIrwin:
This looks great, thanks! Sorry for the slow response.

PR #77, opened 2015-02-24
Fixes #76

### LINKED ISSUE #76: Add support for headers when using nginx
https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Request.php#L34

The function `getallheaders` is not available when using php-fpm. 
https://bugs.php.net/bug.php?id=62596

Maybe the client could define it's own function?
http://php.net/manual/en/function.getallheaders.php#84262

jacobmarshall:
Uses the `getallheaders` function (when available), and falls back to the code snippet @arthens linked to.


ConradIrwin:
Did you test that this works under php-fpm ?

PR #79, opened 2015-03-18
Added check if not mandatory function key actually exists

gallna:
A key 'function' is not mandatory in php stack trace array.

### DIFF
--- a/src/Bugsnag/Stacktrace.php
+++ b/src/Bugsnag/Stacktrace.php
@@ -37,7 +37,12 @@ class Bugsnag_Stacktrace
         // PHP backtrace's are misaligned, we need to shift the file/line down a frame
         foreach ($backtrace as $frame) {
             if (!self::frameInsideBugsnag($frame)) {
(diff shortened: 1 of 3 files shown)


keeganlow:
Hi @gallna, thanks for the pull request. Out of curiosity, what was the situation that gave rise to the need for this change?

gallna:
Include or require is treated as step in stack trace, but function name "include" is not displayed
vendor/laravel/framework/src/Illuminate/View/Engines/PhpEngine.php:39```try{	include $__path;}catch (Exception $e){	$this->handleViewException($e, $obLevel);}```
Spotted this bug when error appeared inside included file.

keeganlow:
Got it. Would you mind adding a test for this?

gallna:
Test added. I discovered, that this bug has been caused by using xdebug_get_function_stack as exception trace.xdebug_get_function_stack is used as exception trace i.a. by Symfony\Component\Debug\Exception\FatalErrorException class.
Test fixture has been taken directly from stack trace caused error in bugsnag.

keeganlow:
Cool! Thanks @gallna.

PR #80, opened 2015-03-19
Added custom curl options support

gallna:
To have the opportunity to add curl options when notification is send.
I want to change default CURLOPT_IPRESOLVE from CURL_IPRESOLVE_WHATEVER to CURL_IPRESOLVE_V4  or CURL_IPRESOLVE_V6 .
It takes 5 sec to send notification when CURLOPT_IPRESOLVE is set to default, change this option reduce time to 0.2 sec.

### DIFF
--- a/src/Bugsnag/Client.php
+++ b/src/Bugsnag/Client.php
@@ -264,6 +264,21 @@ class Bugsnag_Client
     }
 
     /**
+     * Set custom curl options
+     *
+     * @param Array $curlOptions an array with curl options. Eg:
+     *        array(
+     *            CURLOPT_IPRESOLVE => CURL_IPRESOLVE_V4
+     *            )
+     */
+    public function setCurlOptions(array $curlOptions)
+    {
+        $this->config->curlOptions = $curlOptions;
+
+        return $this;
+    }
+
+    /**
      * Set a custom function to call before notifying Bugsnag of an error.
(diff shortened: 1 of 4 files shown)


loopj:
@gallna maybe we should just change our default to CURL_IPRESOLVE_V4?

gallna:
Enough for me, i just didn't dare to ask to change something directly inside Notification class.

keeganlow:
Could you add a test that checks that this behaves correctly, when given a valid curl option?

gallna:
I am not sure what behavior you are expecting to test.

jessicard:
@galina Just having some testing around valid option behavior would be great. For example, a simpe test that checks when a valid option is passed, we don't expect an error. Thanks! :)

Antnee:
Do you mean to test the type is an `array`, or that it's a valid cURL option? AFAIK the only way to check whether options are valid or not would require that a cURL handle be created and either `curl_setopt()` run on each or `curl_setopt_array()` on the array. Both return `false` if an invalid option is given so that could be handled.

Very keen to get this PR merged ASAP though. Nice one @gallna :)

jessicard:
@galina @antnee Thanks for the feedback :) I think we should go ahead and just change the default like @loopj recommended - I made a PR here: https://github.com/bugsnag/bugsnag-php/pull/81

PR #81, opened 2015-03-26
Change default to CURL_IPRESOLVE_V4

jessicard:


Antnee:
@jessicard, this suits our specific requirements perfectly, but are we sure that it will work for everyone? Isn't allowing cURL options to be specified more useful? [Guzzle](https://github.com/guzzle/guzzle/), for example, allows us to specify cURL options by simply [passing in an array](http://guzzle.readthedocs.org/en/latest/faq.html#how-can-i-add-custom-curl-options), much as @gallna did in #80. I'm just concerned that what works for us may break things for others, and it's impossible to know that would be the case with unit tests

jessicard:
@Antnee - good point! I do think we should default it to `CURL_IPRESOLVE_V4` so that the configuration is right for most people, and then I think we should also merge in https://github.com/bugsnag/bugsnag-php/pull/80 so people can specify options. Thanks for your input!

Antnee:
I love it :) Thanks @jessicard

PR #82, opened 2015-03-27
Add example app

jessicard:


loopj:
Nice! we probably don't need a `php` subdirectory underneath the example directory, but overall looks good

jessicard:
@loopj yeah, wasn't sure if we'd be adding more apps for any reason (like our laravel example directory has multiple)

PR #83, opened 2015-04-01
Support PUT payload

forgandenny:


snmaynard:
Looks good, @forgandenny - would you be able to add some tests for the payload parsing of PUTs?

forgandenny:
I may write it at the some point -- as your getRequestMetaData() hasn't got any test at the moment

kattrali:
Merged in 100689a81382dcd4dda1454add0c5edeb2e3b61a, released in [2.7.0](https://github.com/bugsnag/bugsnag-php/blob/master/CHANGELOG.md#270-2016-03-09)

PR #86, opened 2015-05-27
Include class names for class methods in stacktrace

jesseschalken:
- Include the class name in the "method" property for each stack frame.
- Remove "class" property which doesn't exist according to https://bugsnag.com/docs/notifier-api


ConradIrwin:
This looks great, thanks!

PR #91, opened 2015-07-10
Add options not to send $_COOKIE and $_SESSION

jesseschalken:
Adds:

```php
$client = new Bugsnag_Client('...');

$client->setSendCookies(false);
$client->setSendSession(false);
```

Options default to `true` for backwards compatibility.

Also fix some inspection errors in PhpStorm (undefined method, inconsistent return points, incomplete doc comment, etc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
-vendor
-composer.lock
+/vendor
+/composer.lock
+/composer.phar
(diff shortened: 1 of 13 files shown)


snmaynard:
LGTM - Thanks @jesseschalken!

PR #97, opened 2015-09-23
Encode non UTF-8 before send

andtyl:
I have an problem with my project that is encoded in `ISO-8859-1` if the exception message or the code snippet contains non UTF-8 the request fails as `json_encode` returns false on the data.

In this commit https://github.com/bugsnag/bugsnag-php/commit/23a34e1fadfe66bd2578deddb3553531e6c550d5 there was implmented a fix for this (same as mine) but it was rewritten in this commit https://github.com/bugsnag/bugsnag-php/commit/49e5d600014990255d1951fb4c14592d32b70b97 to only affect the metaData array.

What do you think?

### DIFF
--- a/src/Bugsnag/Notification.php
+++ b/src/Bugsnag/Notification.php
@@ -92,6 +92,7 @@ class Bugsnag_Notification
 
     public function postJSON($url, $data)
     {
+        $data = $this->utf8($data);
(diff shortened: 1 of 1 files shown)


andtyl:
Fixed in other PR

PR #98, opened 2015-10-01
Make Configuration and Diagnostics injected to Client through constructor parameters

xhook:
Current implementation has one drawback when it comes to writing custom notification process.

There is a public method `notify(Bugsnag_Error $error, $metaData = array())` which allows to notify using custom constructed `Bugsnag_Error` instance. The only problem is that consumer cannot construct this object correctly because `Bugsnag_Error::_construct` demands passing `Bugsnag_Configuration` instance along with `Bugsnag_Diagnostics` for self initialisation. Instances of these classes are created in constructor of `Bugsang_Client` and kept there private. In the end nevertheless `Bugsang_Client::notify` method being public it is not possible to use it without faking `Bugsnag_Configuration` and `Bugsnag_Diagnostics` instances or extracting them from `Bugsang_Client` using reflection hacks.

This PR fixes this problem by allowing injection of `Bugsnag_Configuration` and `Bugsnag_Diagnostics` as constructor parameters to `Bugsang_Client`. It maintains backward compatibility with previous implementation.

Some topics for further discussion:
1. Should the set* methods be moved to `Bugsnag_Configuration` from `Bugsang_Client`?
2. Should the construction of `Bugsang_Client` be moved to a factory class?

### DIFF
--- a/src/Bugsnag/Client.php
+++ b/src/Bugsnag/Client.php
@@ -3,28 +3,39 @@
 class Bugsnag_Client
 {
     private $config;
+    private $diagnostics;
     /** @var Bugsnag_Notification|null */
     private $notification;
 
     /**
      * Initialize Bugsnag
      *
-     * @param String $apiKey your Bugsnag API key
+     * @param String|Bugsnag_Configuration $config your Bugsnag API key or an instance of Bugsnag_Configuration containing the API key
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR. This is breaking though, so would have to go to our develop branch please.

GrahamCampbell:
Thanks for this. We'll take care of this on our develop branch.

Issue #101, opened 2015-12-03
Support Throwable


GrahamCampbell:
It would be good if notifyException would support Throwable.

kattrali:
This is a great idea, @GrahamCampbell. I'll be looking into making `notifyException` work well with PHP7, as well as some options for preserving backwards compatibility. Any suggestions are also welcome.

cmstone:
It seems that the only way to preserve the compatibility between 5.x and 7.x is to remove the parameter type declaration from the function definition.  7.x Errors and Exceptions implement Throwable interface which has the same methods as the 5.x Exception class

kattrali:
Thanks for the report, @GrahamCampbell. An updated version should be out later today with a fix.

PR #103, opened 2015-12-09
PHP 7 Throwable support:  Error handling type hints to Throwable w/ renamed variables to reflect the change

cmstone:
PHP 7 changes the way Exceptions are thrown.  Errors are thrown now instead of Exceptions for errors (see http://php.net/manual/en/language.errors.php7.php).  Bugsnag was hinted to just for the Exception type, which Error doesn't inherit from.  Both Error and Exception implement the Throwable interface when thrown.


kattrali:
Thank you for the contribution, @cmstone. I looked over the changes and while it works as expected on PHP7, we first need to have a look at backwards compatibility with previous version of PHP to not break the library for users who are unable to immediately upgrade.

I'm open to discussing some options on #101, but can't merge this as-is.

PR #104, opened 2015-12-15
Fix for stacktrace not encoded in UTF-8

GaetanNaulin:
Actually, the stacktrace isn't encoded in UTF-8. When try to send data, json_encode return false and the request return 401 Bad Request.

I think the line 228 can cause encoding problem too.

### DIFF
--- a/src/Bugsnag/Error.php
+++ b/src/Bugsnag/Error.php
@@ -180,7 +180,7 @@ class Bugsnag_Error
         $exceptionArray[] = array(
             'errorClass' => $this->name,
             'message' => $this->message,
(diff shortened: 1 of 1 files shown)


kattrali:
Thank you for the contribution, @GaetanNaulin. The change looks good. No worries about the CI failure, seems to be something infrastructure-related.

GaetanNaulin:
You're welcome ! I'm glad I could contribute something to a project as
great as bugsnag !

2015-12-16 19:19 GMT+01:00 Delisa Mason <notifications@github.com>:

jacobmarshall:
Amen brother; this man speaks the truth!

PR #105, opened 2015-12-18
Fix for error or message not encoded in UTF-8

GaetanNaulin:
Added compatibility with customized errors if they are not encoded in UTF-8.


kattrali:
Ah, cleaning the whole payload makes sense. Looks good, @GaetanNaulin.

PR #107, opened 2015-12-23
Fixed phpunit version

GrahamCampbell:
PHPUnit 5 is the first version to actually support PHP 7.


kattrali:
Thanks, @GrahamCampbell.

PR #111, opened 2016-02-12
Add timeout for fopen

bzick:
Context option timeout added in PHP 5.2.
See http://php.net/manual/en/context.http.php


kattrali:
Thanks! Looks good, @bzick.

PR #112, opened 2016-02-22
Making it possible to delegate plain php errors to the Bugsnag client.

maximkott:
Right now it is not possible to delegate error information to the Bugsnag client. The method **notifyError** lacks several important parameters and leads to information loss. Please consider adding this small feature as soon as possible.

### DIFF
--- a/src/Bugsnag/Client.php
+++ b/src/Bugsnag/Client.php
@@ -480,6 +480,41 @@ class Bugsnag_Client
         $this->notify($error, $metaData);
     }
 
+    /**
+     * Notify Bugsnag of a regular PHP error. This is useful when Bugsnag is not
+     * the "main" error handler and errors are delegated by someone else.
+     *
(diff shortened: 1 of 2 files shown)


kattrali:
Hi @maximkott, thank you for the contribution. It looks like error information is already delegated to the Bugsnag client via the [`errorHandler`](https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Client.php#L496). Can you expand on your use case for why more is needed?

maximkott:
I'm sorry for my late answer, but you are totally right. I don't know how I've missed that. The only benefit you would gain is some consistency, like `notifyException`, `notifyError`. But it is not necessary at all.

kattrali:
Gotcha, thanks @maximkott. I think better interface consistency is something to look into for the next major version, though for now closing this out.

PR #113, opened 2016-02-27
Only filter metadata, not exceptions

duncanhewett:
Currently any filters set using `$bugsnag->setFilters` will filter values that match the key in 'exceptions' in addition to keys in 'metadata'.  This means that the 'code' element can be assigned the string "[FILTERED]".  Although this is currently accepted by Bugsnag, it doesn't meet the schema so may not be in the future.

The correct way to prevent code snippets from being sent is to use the existing `$bugsnag->setSendCode(false)`.


kattrali:
This makes sense. Thanks @duncanhewett.

PR #115, opened 2016-03-13
Fix curl performance

aleemb:
See https://github.com/bugsnag/bugsnag-agent/issues/8 for more. Curl adds an `Expect` header which causes the agent to trip up and cause undue delays.

The other alternative is to force an `HTTP/1.0` connection.


aleemb:
@kattrali any chance of merging this PR? I would like to move to official repo pending this merge.

kattrali:
Sorry about the delay, but this has now been manually merged into core and will be in a patch release shortly. Thanks again!

PR #122, opened 2016-05-17
Add .gitattributes file

vinkla:
This will allow us to exclude files in production with the `--prefer-dist` flag.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+* text=auto
+
+/example export-ignore
+/tests export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/phpunit.xml export-ignore
+/pharbuilder.php
+/CHANGELOG.md export-ignore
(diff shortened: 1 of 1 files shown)


kattrali:
Thank you for the contribution, @vinkla. Somehow I missed this PR when writing the `.gitattributes` file, though there should be one in place now.

vinkla:
Okay, great. Maybe you could use the one in my PR? The current one is broken.

kattrali:
@vinkla How is it broken?

vinkla:
The phpunit line. Why does it only exclude three files?

vinkla:
If you compare the current one with this PR you will see that this PR includes more files.

kattrali:
The example and testing files could conflict with development environments. Not so much the readme and license, et cetera.

vinkla:
Yeah but this doesn't matter for development environment. These files will only be ignored in production. Which is good. Saves space.

GrahamCampbell:
We'll get this sorted for you now. :)

vinkla:
:+1:

PR #127, opened 2016-06-20
Use branch aliases

GrahamCampbell:
Composer supports branch aliases, so someone can ask for `2.7.*@dev`, and get our master branch so they don't have to use the horrible `dev-master` hack.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,5 +23,10 @@
         "psr-0" : {
             "Bugsnag_" : "src/"
         }
-    }
+    },
+    "extra": {
+        "branch-alias": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -23,5 +23,10 @@
         "psr-0" : {
             "Bugsnag_" : "src/"
         }
-    }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "2.7-dev"
+        }
+    },
```
GrahamCampbell: syntax error

GrahamCampbell:
I think travis will be another week on this one...

PR #145, opened 2016-06-21
Implements merging of metadata values in Bugsnag_Client::setMetaData()

### LINKED ISSUE #121: [Feature] setMetaData
Hello,

As we can see in the [error class](https://github.com/bugsnag/bugsnag-php/blob/9bae5c7f8940a390f8c3ef1b7b83f23419a2cd0c/src/Bugsnag/Error.php#L146) when we use `setMetaData()` the existing meta data gets merged with the one passed as an argument.

Is it possible to implement this feature for the client class [here](https://github.com/bugsnag/bugsnag-php/blob/9bae5c7f8940a390f8c3ef1b7b83f23419a2cd0c/src/Bugsnag/Client.php#L273)?

The reason that is useful is because sometimes we want to pass different  meta data in different places of the app. For example a `user_id` at the authentication service and some other value at some other place of the code. Calling `setMetaData()` twice will keep only the latest information.

Thanks for this nice package!

nickma42:
This is intended to fix #121 and also introduces a `getConfig` method for the client which seems like a useful addition and also facilitates testing of the client config.

If exposing the client config is a problem due to the accessibility of the API key is there a better way to facilitate testing of the private config variable?

### DIFF
--- a/src/Bugsnag/Client.php
+++ b/src/Bugsnag/Client.php
@@ -33,6 +33,22 @@ class Bugsnag_Client
     }
 
     /**
+     * Returns client config object or a single property
+     * if provided
+     *
+     * @param string|null $prop
+     * @return mixed
+     */
+    public function getConfig($prop = null)
+    {
+        if ($prop) {
+            return $this->config->get($prop);
+        }
+
+        return $this->config;
+    }
+
+    /**
      * Set your release stage, eg "production" or "development".
      *
      * @param string $releaseStage the app's current release stage
@@ -292,7 +308,7 @@ class Bugsnag_Client
      */
     public function setMetaData(array $metaData)
     {
-        $this->config->metaData = $metaData;
+        $this->config->metaData = array_merge_recursive($this->config->metaData, $metaData);
 
         return $this;
     }
--- a/src/Bugsnag/Configuration.php
+++ b/src/Bugsnag/Configuration.php
@@ -35,7 +35,7 @@ class Bugsnag_Configuration
     public $appVersion;
     public $hostname;
 
-    public $metaData;
+    public $metaData = array();
     public $beforeNotifyFunction;
     public $errorReportingLevel;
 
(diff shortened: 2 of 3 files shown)


src/Bugsnag/Client.php
```diff
@@ -292,7 +308,7 @@ public function setType($type)
      */
     public function setMetaData(array $metaData)
     {
-        $this->config->metaData = $metaData;
+        $this->config->metaData = array_merge_recursive($this->config->metaData, $metaData);
```
GrahamCampbell: We probably shouldn't change this, because now we're unable to remove existing entries. Instead, we should add a new method, perhaps `addMetaData`?

src/Bugsnag/Client.php
```diff
@@ -33,6 +33,22 @@ public function __construct($apiKey)
     }
 
     /**
+     * Returns client config object or a single property
+     * if provided
+     *
+     * @param string|null $prop
+     * @return mixed
+     */
+    public function getConfig($prop = null)
```
GrahamCampbell: I don't think this is needed?

nickma42: @GrahamCampbell that's fair - what'd be the best way to test a new `addMetaData` method without access to the config object?

GrahamCampbell: Just make that function use the same code you modified the setMetaData function too?

GrahamCampbell: And then keep the setMetaData function the same.

nickma42: Sorry, I meant in the unit test to verify the behavior. Is a unit test even necessary given it's just testing `array_recursive_merge` essentially? I was just going by the unit test for the same behavior for the `Error` class.

GrahamCampbell:
I've implemented this slightly differently. Thanks for the PR and the suggestion. :)

PR #162, opened 2016-06-21
Optionally merge the meta data

### LINKED ISSUE #121: [Feature] setMetaData

GrahamCampbell:
Closes #121. Closes #145.

### DIFF
--- a/src/Bugsnag/Client.php
+++ b/src/Bugsnag/Client.php
@@ -313,7 +313,7 @@ class Bugsnag_Client
      */
     public function setMetaData(array $metaData, $merge = false)
     {
-        $this->config->metaData = $merge ? $metaData : array_merge_recursive($this->config->metaData, $metaData);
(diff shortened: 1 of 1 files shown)


kattrali:
I think this makes sense. We'll need to update the docs when this ships.

GrahamCampbell:
Ok, sure. :)

PR #167, opened 2016-06-21
Dropped the phar builder and the old autoloader

GrahamCampbell:


GrahamCampbell:
GitHub Seem to have diffed this wrong.

PR #171, opened 2016-06-22
Finished removing the old ssl code

GrahamCampbell:


GrahamCampbell:
For reference, the first half has done in https://github.com/bugsnag/bugsnag-php/pull/154.

PR #185, opened 2016-06-22
Added upgrading docs

GrahamCampbell:


UPGRADING.md
```diff
@@ -0,0 +1,9 @@
+Upgrading
+=========
+
+Our PHP library has gone through some major improvements, and there are some small changes you'll need to make to get onto the new version.
+
+1. We now require PHP 5.5.9 or higher. If you're using an older version of PHP, you can still use v2. We will containue to maintain v2 along side v3.
+2. We are now using namespaces. Simply replace `Bugsnag_` with `Bugsnag\`, and you're good to go.
+3. We've removed any deprecated functionality. The main thing to watch out for is the removal of the `setUseSSL` function. You should instead just provide URIs that include the scheme. This change will likely only affect enterprise users.
+4. We've changed how our configuration system works. You can now build up our config object in a similar way to how you configured the client in v2, and then pass that as the first paramater when you construct the client object.
```
kattrali: We could link to the example project's configuration here, for reference.

kattrali:
This should maybe include multiple upgrading subheadings, though for now it would just be one for 2.x -> 3.x.

GrahamCampbell:
Or just name the file `UPGRADING-3.0`?

kattrali:
We could, though having more than one or two would be cluttered. What about something similar to the changelog format, like:

```md
# Upgrading

## v2 to v3
…
```

GrahamCampbell:
Something like this then: https://github.com/thephpleague/factory-muffin/blob/2.1/UPGRADING.md?

kattrali:
Yep! That looks good.

GrahamCampbell:
We can include a link to our v3 docs when they're ready. We can write some examples and stuff there.

PR #193, opened 2016-06-22
Refactor to use guzzle

### LINKED ISSUE #146: Refactor to use guzzle

GrahamCampbell:
Closes #146.

### DIFF
--- a/tests/ConfigurationTest.php
+++ b/tests/ConfigurationTest.php
@@ -23,17 +23,6 @@ class ConfigurationTest extends TestCase
         new Configuration([]);
     }
 
-    public function testDefaultEndpoint()
-    {
-        $this->assertSame($this->config->getNotifyEndpoint(), 'https://notify.bugsnag.com');
-    }
-
-    public function testCustomEndpoint()
-    {
-        $this->config->endpoint = 'http://localhost';
-        $this->assertSame($this->config->getNotifyEndpoint(), 'http://localhost');
-    }
-
     public function testDefaultReleaseStageShouldNotify()
     {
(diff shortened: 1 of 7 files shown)


GrahamCampbell:
I'll update the upgrading PR with the changes shortly.

tests/ConfigurationTest.php
```diff
@@ -23,17 +23,6 @@ public function testDoesNotAcceptBadApiKey()
         new Configuration([]);
     }
 
-    public function testDefaultEndpoint()
-    {
-        $this->assertSame($this->config->getNotifyEndpoint(), 'https://notify.bugsnag.com');
```
kattrali: There should probably be an equivalent test to ensure the default guzzle configuration points to the endpoint.

GrahamCampbell: No need. Guzzle already has tests for this. We know it works.

GrahamCampbell: https://github.com/guzzle/guzzle/blob/master/tests/ClientTest.php#L57-L72

GrahamCampbell: User can set the base uri by calling `$client->getGuzzle()->something else i can't remember`.

kattrali: oh, I mean testing that our default instance base uri == `https://notify.bugsnag.com`. A typo in our constant or something should be caught.

GrahamCampbell: Ok, sure. I can add that.

GrahamCampbell: Done.

kattrali:
🎉  Awesome work!

PR #198, opened 2016-06-23
Refactored to use request objects and resolvers

### LINKED ISSUE #99: Make retrieval of server variables pluggable
We're currently using bugsnag on a google app engine application using Laravel.

Unfortunately, the following line is causing us some issues: https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Request.php#L62

Now, given that this isn't bugsnag-laravel, I understand that we can't leverage any of the configuration or request/response libraries.  Although it would be ideal if the mechanism used to obtain values like these was pluggable.  Then, bugsnag-laravel could ship an implementation that retrieves the variables from its own abstraction.  Then, bugsnag-php can ship defaults that use the superglobals.

GrahamCampbell:
Closes #99.


kattrali:
🎉 Looks great.

GrahamCampbell:
:ship:

Issue #96, opened 2015-09-03
Users cant set their own userid in metadata


snmaynard:
This -> https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Error.php#L157 -> can overwrite user id passed through with metadata.

GrahamCampbell:
We don't support setting the user id using the metadata. You need to set the user, specifically, which is different. In some situations, it will work though, due to how bugsnag's backend will process it (ie, when there was no "user" set originally, it'll look at the metadata.

snmaynard:
I dont see how I can configure a user as a part of a notify -> http://docs.bugsnag.com/platforms/php/other/reporting-handled-errors/

GrahamCampbell:
http://docs.bugsnag.com/platforms/php/other/#identifying-users

GrahamCampbell:
;)

GrahamCampbell:
What is the use case for setting it as part of calling notify?

GrahamCampbell:
Surely people will normally be calling bugsnag through a logging abstraction anyway?

GrahamCampbell:
I guess we could support setting the user on the error object on master.

GrahamCampbell:
Here we go @snmaynard: https://github.com/bugsnag/bugsnag-php/pull/209. Sorry about the misunderstanding!

PR #209, opened 2016-06-24
Support completely overriding the user on errors

GrahamCampbell:


kattrali:
Looks good, its a nice fix for #96.

Issue #205, opened 2016-06-24
Separate the client from the actual exception handling registration


GrahamCampbell:
People should be able to use the client without having us messing with their shutdown function and error handling.

PR #213, opened 2016-06-24
Dropped autoNotify and added an exception handler

GrahamCampbell:
People can now just do this:

```php
$client = new Bugnsag\Client(new Bugsnag\Configuration('API-KEY'));

Bugnsag\Handler::register($client);
```

and they'll have our exception handler all set up for them to use natively.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -75,8 +75,6 @@ class Client
         $this->diagnostics = new Diagnostics($this->config, $this->resolver);
         $this->guzzle = $guzzle ?: new Guzzle(['base_uri' => static::ENDPOINT]);
 
-        // Register a shutdown function to check for fatal errors
-        // and flush any buffered errors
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
Closes #205.

src/Client.php
```diff
@@ -74,10 +74,20 @@ public function __construct(Configuration $config, PipelineInterface $pipeline =
...
 
-        // Register a shutdown function to check for fatal errors
-        // and flush any buffered errors
-        register_shutdown_function([$this, 'shutdownHandler']);
+    /**
+     * Destroy the client instance.
+     *
+     * @return void
+     */
+    public function __destruct()
```
GrahamCampbell: I'll revert this back to a shutdown function since php doesn't call destructors when there are fatal errors.

kattrali:
boom 💥

PR #239, opened 2016-06-29
Dynamic config refactor

GrahamCampbell:


src/Configuration.php
```diff
@@ -102,12 +363,87 @@ public function getAppData()
...
+     *
+     * You can use this to add custom tabs of data to each error on your
+     * Bugsnag dashboard.
+     *
+     * @param array[] $metaData an array of arrays of custom data
+     * @param bool    $merge    should we merge the meta data
+     *
+     * @return $this
+     */
+    public function setMetaData(array $metaData, $merge = true)
```
GrahamCampbell: Note the BC break here. Setting merging to be turned on by default to be consistent with the behaviour on the error class.

PR #247, opened 2016-06-30
Allow users to control batch sending

GrahamCampbell:
If someone enables batch sending, it now really does happen. It's no longer conditional on there existing an http request being processed.


GrahamCampbell:
They can call `flush()` to manually send and queued errors whenever they want too.

PR #282, opened 2016-07-06
Don't expose custom user stuff or middlewares

GrahamCampbell:


GrahamCampbell:
Paired with https://github.com/bugsnag/bugsnag-laravel/pull/135.

PR #286, opened 2016-07-07
Rename type to appType

kattrali:
This is in line with the other bugsnag libraries


GrahamCampbell:
Very nice. That name was bugging me actually. Didn't change it because I wasn't sure if that was the standard you wanted. Forgot to ever ask. :)

kattrali:
Always ask :D

PR #295, opened 2016-07-08
Lowered the minimum PHP version

GrahamCampbell:


kattrali:
We should add 5.4 to the build matrix

PR #298, opened 2016-07-08
[WIP] Grouping improvements

GrahamCampbell:
The aim here is to send tokens to Bugsnag so we can more reliably group the code.


GrahamCampbell:
Replaced by #300.

GrahamCampbell:
I didn't like where this was going.

PR #302, opened 2016-07-18
Fallback to localhost when HTTP_HOST is not set

GrahamCampbell:
This is what symfony do when building an HTTP request: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Request.php#L314.


kattrali:
Makes sense, thanks @GrahamCampbell

PR #303, opened 2016-07-18
Don't allow usage on a security compromised version of guzzle

GrahamCampbell:
Patches [CVE-2016-5385](https://httpoxy.org/).


kattrali:
This is a bit outside the bounds of what we should enforce as library maintainers; users should upgrade their dependencies as needed.

PR #305, opened 2016-07-18
Don't hard fail on sending HTTP requests

GrahamCampbell:
This mimics the behaviour in 3.x, and fixes the problem raised by this user: https://github.com/bugsnag/bugsnag-laravel/issues/68#issuecomment-231661059.


kattrali:
Looks good to me

PR #308, opened 2016-07-21
Use composer's ca bundle

GrahamCampbell:
If people are having pear verification issues, all they'll have to do is `composer require composer/ca-bundle`, and it'll start working. This may be particularly of use to Windows users.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "suggest": {
         "composer/ca-bundle": "Composer's CA bundle can be useful if you are having pear verification issues."
-    }
+    },
     "autoload": {
         "psr-4" : {
             "Bugsnag\\" : "src/"
--- a/composer.json
+++ b/composer.json
@@ -12,15 +12,13 @@
     }],
     "require": {
         "php": ">=5.5",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -19,6 +19,9 @@
         "phpunit/phpunit": "^4.8|^5.0",
         "php-mock/php-mock-phpunit": "^1.1"
     },
+    "suggest": {
+        "composer/ca-bundle": "Composer's CA bundle can be useful if you are having pear verification issues."
```
snmaynard: :pear: :trollface:

GrahamCampbell: pear with me...

src/Client.php
```diff
@@ -92,14 +93,31 @@ public function __construct(Configuration $config, ResolverInterface $resolver =
...
     }
 
     /**
+     * Make a new guzzle client instance.
+     *
+     * @param string|null $base
+     *
+     * @return \GuzzleHttp\ClientInterface
+     */
+    protected static function makeGuzle($base = null)
```
GrahamCampbell: Thank you!

PR #317, opened 2016-07-25
Added a way to signify deploys

GrahamCampbell:
Our ruby library has this available, and I think it would be really cool if the PHP one had it too.

It would be really nice to just be able to write `Bugsnag::deploy()` in Laravel, and that would notify Bugsnag of the deploy.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -230,6 +230,34 @@ class Client
     }
 
     /**
+     * Notify Bugsnag of a deployment.
+     *
+     * @param string|null $repository the repository from which you are deploying the code
+     * @param string|null $branch     the source control branch from which you are deploying
+     * @param string|null $revision   the source control revision you are currently deploying
+     *
+     * @return void
+     */
+    public function deploy($repository = null, $branch = null, $revision = null)
+    {
+        $data = [];
+
+        if ($repository) {
+            $data['repository'] = $repository;
+        }
+
+        if ($branch) {
+            $data['branch'] = $branch;
+        }
+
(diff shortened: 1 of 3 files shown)


kattrali:
This is a great idea! Very ergonomic for the framework libraries to use too. 

One concern though is that it probably shouldn’t be in the `Client` since we don’t want people arbitrarily or accidentally sending deploy notifications from applications. Maybe we could have a separate separate `Deploy` class which has a `HttpClient`. Then usage would be something like `Bugsnag::Deploy::send(…)`

GrahamCampbell:
Maybe, but having a separate class here would mean a lot lot more code. The current class is called Client, not Notifier.

GrahamCampbell:
Having this here also has the advantage that when people set the release stage in their bootstrapping process (such as in the boot method in a laravel service provider) all this stuff makes it into the deploy information.

kattrali:
That’s true. It makes sense to do this in a way that allows us to capture as many of the deployment fields as possible without additional overhead. Rolling with it. 🚄

GrahamCampbell:
:D

morloderex:
Curios does this support svn repos which is protected with a username or a password?

GrahamCampbell:
It specifically only supports github.com at the moment.

GrahamCampbell:
We only use the repository url in order to link you to your code from app.bugsnag.com. You can still use the deployment API if you don't use GitHub, just that feature in the dashboard will be unavailable. All other features will still work just the same.

PR #323, opened 2016-08-02
Breadcrumbs

GrahamCampbell:


GrahamCampbell:
Ok, this is ready for review now.

src/Breadcrumbs/Recorder.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+namespace Bugsnag\Breadcrumbs;
+
+use Iterator;
+
+class Recorder implements Iterator
```
GrahamCampbell: It might be worth implementing `Countable` too.

kattrali: Would it make the capacity maintenance easier or something?

GrahamCampbell: Nope, but it would mean that the recorder is natively countable, so we could call `count($recorder)`, and it would return an integer between 0 and 25 representing the number of breadcrumbs we have stored.

kattrali:
This looks great. ⭐ After going over the interface a bit, I think there is one usability enhancement we could make - if a user specifies an invalid breadcrumb type, rather than throwing, we should convert it automatically to the manual type, perhaps with a warning. That way there isn’t a way for it to be missed in production if the breadcrumb isn’t encountered during development. [I opened a similar issue for JS](https://github.com/bugsnag/bugsnag-js/issues/167). This should probably be the rule where there are no enumerated types.

Other than that, I think its ready to go.

GrahamCampbell:
Ok, sure. PHP doesn't have a way to do a "warning" as such. We could just do the error_log thing that we do when a guzzle request fails I guess. I think an exception would be fine though, since it's a usage error rather than an error the user couldn't avoid.

kattrali:
Makes sense, error_log could be a sensible choice. Its a usage error but probably not the kind we should throw an exception for since we can recover for it. It seems plausible to imagine a scenario where some error is happening so you add a breadcrumb with a mispelled type without checking it thoroughly because things are 🔥🔥.

GrahamCampbell:
That said, if someone mistypes a constant there, they'll get a fatal error, so if we only document sending through constants, it can't really go wrong?

GrahamCampbell:
Going to mimic what I did for if the name was too long and auto-fix it within the client, and keep the strict checks in the breadcrum object constructor.

kattrali:
🚀

tillkruss:
Nice work! Will there be a Laravel middleware as well? 😃

GrahamCampbell:
Not a middleware, since we already record the request and user another way. We use breadcrums in laravel to record events fired. We also record logs as breadcrumbs.

PR #327, opened 2016-08-02
When notifying, leave the error as a breadcrumb for next time

GrahamCampbell:
This probably needs some discussion, so opening a PR to the breadcrumbs branch.

### DIFF
--- a/src/Report.php
+++ b/src/Report.php
@@ -5,6 +5,8 @@ namespace Bugsnag;
 use Bugsnag\Breadcrumbs\Breadcrumb;
 use Exception;
 use InvalidArgumentException;
+use ReflectionClass;
+use ReflectionException;
 use Throwable;
 
 class Report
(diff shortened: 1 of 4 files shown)


src/Report.php
```diff
@@ -285,6 +287,20 @@ public function getName()
...
 
     /**
+     * Get the short name.
+     *
+     * @return string
+     */
+    public function getShortName()
+    {
+        try {
+            return (new ReflectionClass($this->name))->getShortName();
```
GrahamCampbell: If an error had a name of `Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException`, then this will give us `TooManyRequestsHttpException`, which is much better for a breadcrumb. If this ends up longer than 30 characters, that's fine too since this will get trimmed later on.

src/Report.php
```diff
@@ -285,6 +287,20 @@ public function getName()
...
+        } catch (ReflectionException $e) {
```
GrahamCampbell: If the name was not for a class, we're just returning the full name.

GrahamCampbell: There's test coverage for both branches.

PR #333, opened 2016-08-03
Attempt to speed up HHVM

GrahamCampbell:
I have run this on master and can confirm it fixes the HUGE slowdown on HHVM.


GrahamCampbell:
Meant to send this to 2.9...

PR #335, opened 2016-08-03
Test on more versions of HHVM

GrahamCampbell:


GrahamCampbell:
I've forwarded ported this temporarily to check the tests work on the newest code too, and everything looks good. See https://travis-ci.org/bugsnag/bugsnag-php/builds/149428493.

PR #337, opened 2016-08-03
Include the error time in the report payload

GrahamCampbell:


src/Report.php
```diff
@@ -155,6 +162,7 @@ public static function fromNamedError(Configuration $config, $name, $message = n
     protected function __construct(Configuration $config)
     {
         $this->config = $config;
+        $this->time = gmdate('Y-m-d\TH:i:s\Z');
```
GrahamCampbell: Do we need an integer timestamp here, or a formatted timestamp?

kattrali: Formatted timestamp

GrahamCampbell: Ok, I asked after seeing the event worker code treating it as an integer.

kattrali:
⭐ 🚢

PR #351, opened 2016-08-15
Added a getter for the stacktrace

### LINKED ISSUE #331: Add support for accessing exception data on report
Exception data is currently inaccessible to the user and would be useful for customizing the error grouping, such as removing stacktrace lines or using the file or line number as a part of a customized algorithm for grouping hash.

GrahamCampbell:
It is still not possible to set a stacktrace from the outside, since the `setStacktrace` method is protected. It is possible to mutate the stacktrace after calling the new `getStacktrace` method, however we have no easy way to insert frames into anywhere other than the very end of the array, and no way to remove frames once they've already made it through the add frame method.

I don't know if we need to add this other frame manipulation ability, or what it would look like.

---

Closes #331.

### DIFF
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -2,6 +2,7 @@
 
 namespace Bugsnag;
 
+use InvalidArgumentException;
 use RuntimeException;
 use SplFileObject;
 
@@ -134,9 +135,21 @@ class Stacktrace
     /**
      * Get the array representation.
      *
-     * @return array
+     * @return array[]
      */
     public function toArray()
+    {
+        return $this->frames;
+    }
+
+    /**
+     * Get the stacktrace frames.
+     *
+     * This is the same as calling toArray.
+     *
+     * @return array[]
+     */
+    public function getFrames()
     {
         return $this->frames;
     }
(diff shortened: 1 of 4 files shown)


kattrali:
Cool, this looks good. It seems like `Stacktrace` would also need an in-place `filter` or something to meet the constraints of #331, since the underlying problem is that sometimes frames need to be removed.

GrahamCampbell:
Ability to remove frames now added.

src/Stacktrace.php
```diff
@@ -134,14 +135,26 @@ public function __construct(Configuration $config)
...
      */
     public function toArray()
     {
         return $this->frames;
     }
 
     /**
+     * Get the stacktrace frames.
+     *
+     * This is the same as calling toArray.
```
kattrali: lol.

GrahamCampbell: thought it would be nice to clarify they were the same from the outside perspective :)

kattrali:
This is great. The preferred way of editing a report’s stacktrace is now to `getStacktrace()`, `getFrames()` to inspect frames if needed, and `removeFrame(num)` if trimming is needed.

GrahamCampbell:
Yeh. :)

snmaynard:
Is this documented on docs.bugsnag.com? I don't see if but I might be looking in the wrong place.

GrahamCampbell:
I think we intentionally omitted all documentation for the stacktrace class since it was considered internal. Now that we've added (well reintroduced in a better way from v2) the ability to modify the frames, we probably should consider adding a dedicated page covering this stuff.

---

I wonder if a good way to approach this could actually be to have a page explaining the internals of the library for advanced users, and within that guide, explain how to interface with the frames. It could also potentially make a good blog post.

snmaynard:
I think it would presumably appear on the https://docs.bugsnag.com/platforms/php/other/customizing-error-reports/ page with an example of removing logging lines from the stacktrace?

GrahamCampbell:
I see delisa has added some docs there. :)

snmaynard:
@kattrali is on a mission! Nice one Delisa :D!

GrahamCampbell:
👍 🚀

PR #363, opened 2016-09-05
Support fallback app types

GrahamCampbell:


src/Configuration.php
```diff
@@ -353,13 +360,29 @@ public function setAppType($type)
...
+
+    /**
      * Get the application data.
      *
      * @return array
      */
     public function getAppData()
     {
-        return array_merge(['type' => php_sapi_name(), 'releaseStage' => 'production'], array_filter($this->appData));
+        return array_merge(['type' => $this->fallbackType ?: php_sapi_name(), 'releaseStage' => 'production'], array_filter($this->appData));
```
snmaynard: Is this how defaults are set?

GrahamCampbell: Yeh, so the array_filter means that if the user hasn't set an app type, then the stuff from the LHS of the array merge will replace it, and the ternary part just means we use the sapi name if there's no library set fallback.

snmaynard: Why dont we use the constructor to do this?

GrahamCampbell: We could use the constructor to initialize the fallback type if you would prefer that, yeh.

GrahamCampbell: But, if the library unsets the fallback type, then the behaviour would change. The tests would fail if we moved this to the constructor.

GrahamCampbell: Specifically, line 87 of the config test would fail.

snmaynard: That would seem good - that way all the defaults are centralised and we dont have to worry about merging things in the right order so much. We probably still need the fallbackType from what you were saying because we dont know during construction what app we are running in all the time, but it limits the complexity a bit I think

GrahamCampbell: I don't think we should get rid of the array merge. That would be a BC break and would break more than one of our tests.

GrahamCampbell: array_merge is a common way PHP libraries set defaults. Laravel's config system works exactly like this.

snmaynard: If you call setFallbackType with null it could fall back to the default, though I think the user has said null so I would say the test is wrong.

GrahamCampbell: We use `null` across this library to indicate "use the defaults" rather than "unset the value".

GrahamCampbell: That's also something that's a convention (across some - thanks PHP) in PHP's internal functions when you want to use a default or skip args, you pass null, since PHP doesn't support named parameter passing.

kattrali:
Makes sense to me. its basically the way overriding values is handled in cocoa as well.

PR #368, opened 2016-09-21
Don'y define variables within files to be required

GrahamCampbell:
This both fixes a bug, and corrects a bad practice.

The bug fix is the 2nd time this script is run, the bugsnag variable doesn't get defined since php says it's already loaded that file before. The bad practice is the variable definition within the file being required. It's conventional to just return from such files and define the result in your own variable.

### DIFF
--- a/example/php-expanded/index.php
+++ b/example/php-expanded/index.php
@@ -1,10 +1,7 @@
 <?php
 
-require_once 'vendor/autoload.php';
-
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Though, tbh, I'd not recommend having this runtime file. Nobody really does this in php. It's common place to see the vendor autoload file just stuck at the top of the entry point.

PR #372, opened 2016-12-05
Fix legacy PHP docs link

duncanhewett:
Also change links from http to https

### DIFF
--- a/tests/fixtures/code/File.php
+++ b/tests/fixtures/code/File.php
@@ -13,7 +13,6 @@ $exception = new Exception('Exception!');
 
 // Throw the exception
 throw $exception;
-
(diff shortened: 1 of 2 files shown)


tests/fixtures/code/File.php
```diff
@@ -13,7 +13,6 @@
 
 // Throw the exception
 throw $exception;
-
```
GrahamCampbell: Change here is due to a recent change to the StyleCI presets. Config needs updating, then this change reverting so the tests pass again.

GrahamCampbell: See https://github.com/bugsnag/bugsnag-php/pull/373. :)

PR #374, opened 2016-12-06
Allow configuring strip path regex

rohitpaulk:
We use paths that change with every release for our app. For example - `/opt/releases/app-name/201611110115/__src/Class.php`. 

This messes with the filename based grouping that happens for exceptions - We end up getting duplicate reports for the same exception on every new release. 

I saw that one way to prevent this is to set the strip path to the exact value (`/opt/releases/app-name/2016.../`).  However, it'd be easier if I could use regexes here. Seeing that internally a regex is used, this PR makes that value configurable. It doesn't mess with the original behavior.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -251,6 +251,22 @@ class Configuration
     }
 
     /**
+     * Set the regular expression used to strip paths from stacktraces.
+     *
+     * @param string|null $stripPathRegex
+     *
+     * @return void
+     */
+    public function setStripPathRegex($stripPathRegex)
+    {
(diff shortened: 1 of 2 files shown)


rohitpaulk:
Styling issues fixed.

GrahamCampbell:
Thank you @rohitpaulk. Looks good to me. :+1:

rohitpaulk:
Going to rebase on master...

rohitpaulk:
Access to the old repo is messed up, I've re-opened this in https://github.com/bugsnag/bugsnag-php/pull/390

cc: @GrahamCampbell

PR #375, opened 2016-12-08
Reverted bad change

GrahamCampbell:


GrahamCampbell:
Really important, given this broke the tests.

GrahamCampbell:
(well, not important for consumers of the library, but you know what i mean)

PR #376, opened 2016-12-27
Added additional configuration options to deploy function

jmherbst:
Hello, 

I have a deployment application that is responsible for deploying multiple applications that all have related bugsnag projects.  I was going to use this deploy method to send those events to bugsnag for the respective applications upon deploy time, but needed to add additional parameters + overrides in `HttpClient::deploy` for those data keys.  (https://docs.bugsnag.com/api/deploy-tracking/#api-overview)

This does not change any current functionality of the `Bugsnag::deploy()` function.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -288,8 +288,8 @@ class Client
      * @param string|null $repository    the repository from which you are deploying the code
      * @param string|null $branch        the source control branch from which you are deploying
      * @param string|null $revision      the source control revision you are currently deploying
-     * @param string|null $releaseStage the release stage (eg, production, staging) currently being deployed
-     * @param string|null $appVersion   the source control revision id for the code you are deploying
+     * @param string|null $releaseStage  the release stage (eg, production, staging) currently being deployed
+     * @param string|null $appVersion    the source control revision id for the code you are deploying
      * @param string|null $provider      the name of your source control provider
      *
      * @return void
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Thanks for the PR, though, I think it's maybe best for you to call the API yourself for that use case, since this method was designed for an application to notify a deployment of itself. I think it's confusing to allow these settings to be overwritten at runtime.

This package is meant to be a notifier, rather than an "API Client". I'm just concerned about blurring the two. I know this issue (about doing more than just strictly error notifying) was also raised when the deploy method was added to the notifier in the first place.

Don't close the PR though, since it's ultimately not up to me. That was just my initial reaction, and why I specifically didn't add this originally (since those are read from config). :)

// cc @kattrali.

jmherbst:
Thanks for the input and quick response @GrahamCampbell 👍 

I can certainly just make an api call direct.  I posed this question to @snmaynard via bugsnag support the other day too, so I'll tag him here to see what his thoughts are.

GrahamCampbell:
Ok. If the team do want to accept this, note that a camelCase code style is enforced for variable names. Also, note that having a lot of parameters on a function can make it hard to use, especially if you're not in an IDE. Just, some other things to consider.

Cawllec:
Hi @jmherbst, we're going to be reviewing and changing the deployment API relatively soon, and will update the deployment methodology when ready.  Thanks for the contribution

jmherbst:
Sounds good, thanks @Cawllec

PR #381, opened 2017-04-25
Make constructor public instead of protected

Vmak11:
This is needed when you want to call handler methods explicitly without registering default php error/exception/shutdown handlers.  An example use case would be if you utilized your own error/exception/shutdown handlers and within that handler called these handler methods.  You may also call other handler methods unrelated to bugsnag.

### DIFF
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -36,7 +36,7 @@ class Handler
      *
      * @return void
      */
-    protected function __construct(Client $client)
+    public function __construct(Client $client)
     {
         $this->client = $client;
     }


GrahamCampbell:
This class is not designed to be constructed in this way. Which method in particular did you want to call please?

Vmak11:
All of them.  In our previous version of Bugsnag package these methods were available directly from the bugsnag client object and wasn't an issue.  For instance, lets just focus on the error handler method "errorHandler" as an example.  This is a very basic example of my previous explained use case but hopefully gets the point across.  Sorry for any syntax errors in advance.

// just a class that has a custom method for error handling
class CustomErrorClass
{
    protected $handlers;

    public function __construct()
    {
        $this->handlers = [];               
    }

    public function addHandler(AbstractHandler $handler)
    {
        $this->handlers[] = $handler;
    }

    public function customErrorHandler($level, $message, $file, $line)
    {            
        foreach ($this->handlers as $handler) {
            $handler->handleError($level, $message, $file, $line);
        }
    } 
}

// extends AbstractHandler to ensure it has a handleError method
class CustomBugsnagHandler extends AbstractHandler
{
    public function handleError($level, $message, $file, $line)
    {
        $client = Bugsnag\Client::make('YOUR-API-KEY-HERE');
        $handler = new Bugsnag\Handler($client);
        $handler->errorHandler($level, $message, $file, $line);
    }
}

$custom_error_class = new CustomErrorClass();

// A way to add various different error handlers
$custom_error_class->addHandler(new CustomBugsnagHandler());
$custom_error_class->addHandler(new CustomLogHandler());
$custom_error_class->addHandler(new OtherCustomErrorHandler());

// Set your custom error handler, we don't want this overridden by BugSnag register method on Handler class
set_error_handler([$custom_error_class, "customErrorHandler"]);

GrahamCampbell:
Nice idea. I am aware that this class is non-final, so technically this could be regarded as a breaking change: https://3v4l.org/pILQK.

If @kattrali decides that this is an acceptable break, then this PR has a 👍 from me. The use case seems fair.

PR #384, opened 2017-06-06
Fix setStripPath and setProjectRoot regexes

mtibben:
`setStripPath` and `setProjectRoot` has some unexpected behaviour.

For example if you `->setStripPath('/dir1/')` or `->setProjectRoot('/dir1/')` then a path like `/dir1/dir2/dir1/file.php` is stripped to `dir2file.php` - the regex strips all occurrences of `/dir1/` rather than just the root path (which I assume is the intended purpose).

So this updates the `setStripPath` and `setProjectRoot` functions to always match the start of a path string.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -247,7 +247,7 @@ class Configuration
      */
     public function setStripPath($stripPath)
     {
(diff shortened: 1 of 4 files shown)


loopj:
Thanks @mtibben - good catch!

mtibben:
No worries! Any chance of a release to make updating easy?

loopj:
v3.5.1 now live

PR #389, opened 2017-07-09
Fixed callback bug

GrahamCampbell:
Callbacks directly passed to notify must also be able to cancel sending by returning `false`.


tests/ClientTest.php
```diff
@@ -131,6 +131,8 @@ public function testBeforeNotifySkipsError()
     {
         $this->client = new Client($this->config = new Configuration('example-api-key'), null, $this->guzzle);
 
+        $this->client->setBatchSending(false);
```
GrahamCampbell: Without this, the original test would have still passed, even if we didn't skip sending!

PR #393, opened 2017-07-26
Added latest php and hhvm

GrahamCampbell:
Also, since a few days ago, the following is no longer needed for HHVM:

```yaml
    sudo: required
    dist: trusty
    group: edge
```

I'm not sure if you want to keep all these older HHVM versions either?


GrahamCampbell:
For reference, travis are slowly changing the defaults for repos, and who knows which repo changes when. Turns out trusty is not yet the default on this repo yet, but was on my test repo.

PR #395, opened 2017-08-02
Call previously defined exception and error handlers

### LINKED ISSUE #392: Call to set_exception_handler ignores previously set callbacks
My app leverages a custom exception handler. It is registered early on, before the bugsnag handler. When I register the bugsnag handler, it overwrites my custom exception handler.

Because `set_exception_handler` [returns the previously set callback][2], it is possible to create a nice stack of these exception handling callbacks. It would be nice if `Bugsnag\Handler::register` respected this pattern. One example implementation of the expected behavior is [`Monolog\ErrorHandler`][1].

`set_error_handler`  also returns the previously set callback.

[1]: https://github.com/Seldaek/monolog/blob/master/src/Monolog/ErrorHandler.php#L77-L91
[2]: http://php.net/manual/en/function.set-exception-handler.php#refsect1-function.set-exception-handler-returnvalues

mterwill:
This library overwrites previously set user exception and error handlers. Both `set_error_handler` and `set_exception_handler` will return callables for the previously defined handlers if they were set. Switch the library's behavior to respect this.

Closes #392

### DIFF
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -12,21 +12,21 @@ class Handler
     protected $client;
 
     /**
-     * The previously registered error handler returned by the interpreter.
+     * The previously registered error handler.
      *
      * @var callable|null
      */
     protected $previousErrorHandler;
 
     /**
-     * The previously registered exception handler returned by the interpreter.
+     * The previously registered exception handler.
      *
      * @var callable|null
      */
     protected $previousExceptionHandler;
 
     /**
-     * Register our exception handler.
+     * Register our handlers.
      *
      * @param \Bugsnag\Client|string|null $client client instance or api key
      *
@@ -42,7 +42,7 @@ class Handler
     }
 
     /**
-     * Register our exception handler and preserve those previously registered.
+     * Register our handlers and preserve those previously registered.
      *
      * @param \Bugsnag\Client|string|null $client client instance or api key
      *
@@ -88,7 +88,7 @@ class Handler
     }
 
     /**
-     * Register the bugsnag error handler and save the returned value.
+     * Register the bugsnag exception handler and save the returned value.
      *
      * @param bool $callPrevious whether or not to call the previous handler
      *
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -78,7 +78,7 @@ class Handler
      *
      * @return void
      */
-    public function registerErrorHandler($callPrevious = true)
+    public function registerErrorHandler($callPrevious)
     {
         $previous = set_error_handler([$this, 'errorHandler']);
 
@@ -94,7 +94,7 @@ class Handler
      *
      * @return void
      */
-    public function registerExceptionHandler($callPrevious = true)
+    public function registerExceptionHandler($callPrevious)
     {
         $previous = set_exception_handler([$this, 'exceptionHandler']);
 
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -144,7 +144,7 @@ class Handler
         }
 
         if ($this->previousErrorHandler) {
-            call_user_func(
+            return call_user_func(
                 $this->previousErrorHandler,
                 $errno,
                 $errstr,
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -12,21 +12,82 @@ class Handler
     protected $client;
 
     /**
+     * The previously registered error handler as returned by the interpreter.
+     *
+     * @var callable|null
+     */
+    protected $previousErrorHandler;
+
+    /**
+     * The previously registered exception handler as returned by the interpreter.
+     *
+     * @var callable|null
+     */
+    protected $previousExceptionHandler;
+
+    /**
      * Register our exception handler.
      *
      * @param \Bugsnag\Client|string|null $client client instance or api key
+     * @param bool $callPrevious whether or not to call the previous handlers
      *
      * @return static
      */
-    public static function register($client = null)
+    public static function register($client = null, $callPrevious = true)
     {
         $handler = new static($client instanceof Client ? $client : Client::make($client));
 
-        set_error_handler([$handler, 'errorHandler']);
-        set_exception_handler([$handler, 'exceptionHandler']);
-        register_shutdown_function([$handler, 'shutdownHandler']);
+        $handler->registerErrorHandler($callPrevious);
+        $handler->registerExceptionHandler($callPrevious);
+        $handler->registerShutdownHandler();
 
         return $handler;
+    }
+
+    /**
+     * Register the bugsnag error handler and save the previous one
+     * (if it exists) to call later.
+     *
+     * @param bool $callPrevious whether or not to call the previous handler
+     *
+     * @return void
+     */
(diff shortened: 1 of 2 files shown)


src/Handler.php
```diff
@@ -69,15 +137,23 @@ public function exceptionHandler($throwable)
      */
     public function errorHandler($errno, $errstr, $errfile = '', $errline = 0)
     {
-        if ($this->client->shouldIgnoreErrorCode($errno)) {
-            return;
-        }
```
mterwill: FWIW, the interpreter interprets this `null` return value differently than the `false` you were returning below.

> It is important to remember that the standard PHP error handler is completely bypassed for the error types specified by error_types unless the callback function returns FALSE.
>
> http://php.net/manual/en/function.set-error-handler.php

We can verify this:

```
$ cat test-exits-null.php
<?php

set_error_handler(
    function ($errno, $errstr) {
        return;
    }
);

trigger_error("Uh oh!", E_USER_WARNING);
$ php test-exits-null.php
$ # warning not outputted
```

```
$ cat test-exits-false.php
<?php

set_error_handler(
    function ($errno, $errstr) {
        return false;
    }
);

trigger_error("Uh oh!", E_USER_WARNING);
$ php test-exits-false.php
PHP Warning:  Uh oh! in /private/tmp/test-exits-false.php on line 9

Warning: Uh oh! in /private/tmp/test-exits-false.php on line 9
```

src/Handler.php
```diff
@@ -69,15 +137,23 @@ public function exceptionHandler($throwable)
...
 
-        $report = Report::fromPHPError($this->client->getConfig(), $errno, $errstr, $errfile, $errline);
-
-        $this->client->notify($report);
+            $this->client->notify($report);
+        }
 
-        return false;
+        if ($this->previousErrorHandler) {
+            call_user_func(
```
kattrali: Should this call be returned?

kattrali:
This works well. I left a comment for the one question I had, around returning the value of the previous handler, in the case it will return `false`.

src/Handler.php
```diff
@@ -12,24 +12,85 @@ class Handler
...
 
     /**
+     * Register the bugsnag error handler and save the previous one
+     * (if it exists) to call later.
+     *
+     * @param bool $callPrevious whether or not to call the previous handler
+     *
+     * @return void
+     */
+    public function registerErrorHandler($callPrevious = true)
```
GrahamCampbell: Setting this default to true is a backwards compatibility breaker. It should be set to false by default I think?

src/Handler.php
```diff
@@ -12,24 +12,85 @@ class Handler
...
+    /**
      * Register our exception handler.
      *
      * @param \Bugsnag\Client|string|null $client client instance or api key
+     * @param bool $callPrevious whether or not to call the previous handlers
      *
      * @return static
      */
-    public static function register($client = null)
+    public static function register($client = null, $callPrevious = true)
```
GrahamCampbell: We should be very careful adding new parameters to public or protected methods in PHP since it's technically a BC break. A way around this would be to define a new method called `registerWithPrevious`.

GrahamCampbell: Not only that, but it's more readable. Boolean parameters can be confusing.

src/Handler.php
```diff
@@ -12,24 +12,85 @@ class Handler
...
+        $previous = set_error_handler([$this, 'errorHandler']);
+
+        if ($callPrevious) {
+            $this->previousErrorHandler = $previous;
+        }
+    }
+
+    /**
+     * Register the bugsnag exception handler and save the previous one
+     * (if it exists) to call later.
```
GrahamCampbell: Please ensure that "short descriptions" are less than 8 chars and don't overflow past one line.

src/Handler.php
mterwill: Up to y'all. Setting the default to false will preserve the current broken behavior for new installations and create a bug that can lurk for a while.

This change will only break backwards compatibility people who

1. load their custom handler
1. call `Bugsnag::register`
1. re-register their custom handler on top of that, saving the values returned by `set_x_handler` and re-calling that

I'm guessing those who have already worked around this will have just called `Bugsnag\Handler::xHandler` directly, in which case `previousXHandler` will be `null` since `register` is never called.

I'd think calling the user's function twice is better than never. 🤷‍♂️

GrahamCampbell: > Setting the default to false will preserve the current broken behavior for new installations and create a bug that can lurk for a while.

I don't think the current behaviour is broken actually. The handler is intentionally replacing existing ones. It's a feature if we want to have one that doesn't behave like this.

src/Handler.php
mterwill: Why would optional parameters that preserve the existing behavior break backwards compatibility? [obviously this does not preserve existing behavior, but we can continue that discussion above]

I'm generally not the biggest fan of boolean flags either, but in this case I think removing it would either (a) add a lot of duplicated code or (b) add unnecessary complexity to the existing code.

GrahamCampbell: > Why would optional parameters that preserve the existing behavior break backwards compatibility? [obviously this does not preserve existing behavior, but we can continue that discussion above]

Because PHP allows you to call methods with too many parameters without errors, and this class is non-final so could have been extended with a different signature for the 2nd parameter. Given the current handler didn't support preserving the previous handler, I'm sure other people will have extended it in their projects to add this.

GrahamCampbell: > I'm generally not the biggest fan of boolean flags either, but in this case I think removing it would either (a) add a lot of duplicated code or (b) add unnecessary complexity to the existing code.

I'm only suggesting we don't have it in the register function. The others are fine. :)

src/Handler.php
```diff
@@ -12,6 +12,20 @@ class Handler
...
 
     /**
+     * The previously registered error handler returned by the interpreter.
+     *
+     * @var callable|null
+     */
+    protected $previousErrorHandler;
+
+    /**
+     * The previously registered exception handler returned by the interpreter.
```
GrahamCampbell: I think `The previously registered exception handler.` is probably fine, and similar for above. While the main implementation of the PHP language is an interpreter (after a compilation phase to an intermediate language), there exist other implementations of the language, so no reason to assume the runtime is actually an interpreter. Just something very minor. :P

mterwill:
This should be fully backwards compatible now, with a new `Handler::registerWithPrevious` method added to register bugsnag handlers and preserve previously registered exception handlers.

Your [documentation page][1] should probably be updated to include a note about this once it's released.

[1]: https://docs.bugsnag.com/platforms/php/other/#reporting-unhandled-exceptions

src/Handler.php
GrahamCampbell: > FWIW, the interpreter interprets this null return value differently than the false you were returning below.

Looks like a bug. Either our doc was wrong, and we meant `bool|null`, or we meant to `return false`.

kattrali:
Thanks @mterwill 🎉

PR #398, opened 2017-08-24
Cawllec/console request

Cawllec:
Re [this issue in bugsnag-laravel](https://github.com/bugsnag/bugsnag-laravel/issues/211) this implements a basic console-request class that will take an array of the console commands and attach them as meta-data.

This gives a handler for framework implementations, while leaving it framework-specific on how console input is acquired.

### DIFF
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -11,15 +11,20 @@ class BasicResolver implements ResolverInterface
      */
     public function resolve()
     {
-        if (!isset($_SERVER['REQUEST_METHOD'])) {
-            if (PHP_SAPI ==='cli' && isset($_SERVER['argv'])) {
-                return new ConsoleRequest($_SERVER['argv']);
-            } else {
-                return new NullRequest();
-            }
+        if (isset($_SERVER['REQUEST_METHOD'])) {
+            return new PhpRequest($_SERVER,
+                empty($_SESSION) ? [] : $_SESSION,
+                empty($_COOKIE) ? [] : $_COOKIE,
+                static::getRequestHeaders($_SERVER),
+                static::getInputParams($_SERVER, $_POST));
         }
 
-        return new PhpRequest($_SERVER, empty($_SESSION) ? [] : $_SESSION, empty($_COOKIE) ? [] : $_COOKIE, static::getRequestHeaders($_SERVER), static::getInputParams($_SERVER, $_POST));
+        if (PHP_SAPI === 'cli' && isset($_SERVER['argv'])) {
+            return new ConsoleRequest($_SERVER['argv']);
+        }
+        
+        return new NullRequest();
+
     }
 
     /**
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -12,7 +12,7 @@ class BasicResolver implements ResolverInterface
     public function resolve()
     {
         if (!isset($_SERVER['REQUEST_METHOD'])) {
-            if ((php_sapi_name() == 'cli') && (isset($_SERVER['argv']))) {
+            if ((PHP_SAPI == 'cli') && (isset($_SERVER['argv']))) {
                 return new ConsoleRequest($_SERVER['argv']);
             } else {
                 return new NullRequest();
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -12,7 +12,11 @@ class BasicResolver implements ResolverInterface
     public function resolve()
     {
         if (!isset($_SERVER['REQUEST_METHOD'])) {
-            return new NullRequest();
+            if ((php_sapi_name() == 'cli') && (isset($_SERVER['argv']))) {
+                return new ConsoleRequest($_SERVER['argv']);
+            } else {
+                return new NullRequest();
+            }
         }
 
         return new PhpRequest($_SERVER, empty($_SESSION) ? [] : $_SESSION, empty($_COOKIE) ? [] : $_COOKIE, static::getRequestHeaders($_SERVER), static::getInputParams($_SERVER, $_POST));
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -12,7 +12,7 @@ class BasicResolver implements ResolverInterface
     public function resolve()
     {
         if (!isset($_SERVER['REQUEST_METHOD'])) {
-            if ((PHP_SAPI == 'cli') && (isset($_SERVER['argv']))) {
+            if (PHP_SAPI ==='cli' && isset($_SERVER['argv'])) {
                 return new ConsoleRequest($_SERVER['argv']);
             } else {
                 return new NullRequest();
--- a/src/Request/ConsoleRequest.php
+++ b/src/Request/ConsoleRequest.php
@@ -13,10 +13,12 @@ class ConsoleRequest implements RequestInterface
 
     /**
      * Create a new console request instance.
+     * 
+     * @param array     $commandArray An array of the command line input
      *
      * @return void
      */
-    public function __construct($commandArray)
+    public function __construct(array $commandArray)
     {
         $this->command = $commandArray;
     }
--- a/src/Request/ConsoleRequest.php
+++ b/src/Request/ConsoleRequest.php
@@ -13,7 +13,7 @@ class ConsoleRequest implements RequestInterface
 
     /**
      * Create a new console request instance.
-     * 
+     *
      * @param array     $commandArray An array of the command line input
      *
      * @return void
--- a/src/Request/ConsoleRequest.php
+++ b/src/Request/ConsoleRequest.php
@@ -7,20 +7,20 @@ class ConsoleRequest implements RequestInterface
     /**
      * The unformated console command.
      *
-     * @var array
+     * @var string[]
      */
     protected $command;
 
     /**
      * Create a new console request instance.
      *
-     * @param array     $commandArray An array of the command line input
+     * @param string[] $command an array of the console command input
      *
      * @return void
      */
-    public function __construct(array $commandArray)
+    public function __construct(array $command)
     {
-        $this->command = $commandArray;
+        $this->command = $command;
     }
 
     /**
@@ -30,7 +30,7 @@ class ConsoleRequest implements RequestInterface
      */
     public function isRequest()
     {
-        return true;
+        return false;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
...
+     * The unformated console command.
+     *
+     * @var array
+     */
+    protected $command;
+
+    /**
+     * Create a new console request instance.
+     *
+     * @param array     $commandArray An array of the command line input
```
GrahamCampbell: Alignment bug. :)

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+namespace Bugsnag\Request;
+
+class ConsoleRequest implements RequestInterface
+{
+    /**
+     * The unformated console command.
+     *
+     * @var array
```
GrahamCampbell: `string[]` I think is the type here?

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
...
+    protected $command;
+
+    /**
+     * Create a new console request instance.
+     *
+     * @param array     $commandArray An array of the command line input
+     *
+     * @return void
+     */
+    public function __construct(array $commandArray)
```
GrahamCampbell: Conventional to match the internal property name, so name the param `$command`?

src/Request/BasicResolver.php
```diff
@@ -12,7 +12,11 @@ class BasicResolver implements ResolverInterface
     public function resolve()
     {
         if (!isset($_SERVER['REQUEST_METHOD'])) {
-            return new NullRequest();
+            if ((PHP_SAPI == 'cli') && (isset($_SERVER['argv']))) {
```
GrahamCampbell: The extra brackets are not needed, and `===` should be used for the string comparison.

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Get the request context.
+     *
+     * @return string|null
+     */
+    public function getContext()
+    {
+    }
```
GrahamCampbell: `return implode(' ', $this->command)` perhaps would be a good context, or perhaps, limit this to only the first upto 2 parts of the array?

GrahamCampbell:
Reviewed this in some detail. :)

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Are we currently processing a request?
+     *
+     * @return bool
+     */
+    public function isRequest()
+    {
+        return true;
```
GrahamCampbell: This probably has to be `false`, otherwise, this pull request would be considered a breaking change. Previously, "request" meant "HTTP request", not just "invocation context", or whatever name we wanna give to this. :P

src/Request/ConsoleRequest.php
kattrali: Is this something StyleCI could lint for?

src/Request/ConsoleRequest.php
kattrali: The first two parts would be good, running this with `artisan` commands as well as plain `php` produces useful results, moreso than the default which is the top line of the stacktrace.

src/Request/ConsoleRequest.php
GrahamCampbell: Not yet, but it is planned. I think it's actually been planned for a very long time, since I think you mentioned this feature last year too. :P

src/Request/ConsoleRequest.php
GrahamCampbell: This code will return at most the first 2 parts of the command, concatenated together with a space. If the array is smaller than 2 elements, it is tolerant to that, and will just return the singleton array, or the empty array.

```php
return implode(' ', array_slice($this->command, 0, 2));
```

src/Request/ConsoleRequest.php
Cawllec: They're allegedly the same, but `string[]` is more explicit to the expected array elements

src/Request/ConsoleRequest.php
Cawllec: I agree, to avoid potential issues :)

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Get the request formatted as meta data.
+     *
+     * @return array
+     */
+    public function getMetaData()
+    {
+        if (is_array($this->command)) {
```
GrahamCampbell: No need for the `is_array` check. It is guaranteed to be true. PHP doesn't allow `null` through an `array` type.

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,102 @@
...
+            $commandString = implode(' ', $this->command);
+            $primaryCommand = $this->command[0];
+            $arguments = [];
+            $options = [];
+            foreach (array_slice($this->command, 1) as $arg) {
+                $arg[0] == '-' ? $options[] = $arg : $arguments[] = $arg;
+            }
+            $data = [
+                'Input' => $commandString,
+                'Command' => $primaryCommand,
```
GrahamCampbell: It'd argue this should be `'Command' => $this->getContext(),`, and the above be `foreach (array_slice($this->command, 2) as $arg) {`, otherwise, the command will always be `'php'` (I think?).

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function getMetaData()
+    {
+        if (is_array($this->command)) {
+            $commandString = implode(' ', $this->command);
+            $primaryCommand = $this->command[0];
+            $arguments = [];
+            $options = [];
+            foreach (array_slice($this->command, 1) as $arg) {
+                $arg[0] == '-' ? $options[] = $arg : $arguments[] = $arg;
```
GrahamCampbell: Assignment inside ternaries is somewhat unusual. Perhaps consider just a plain old if. :)

src/Request/BasicResolver.php
```diff
@@ -12,7 +12,11 @@ class BasicResolver implements ResolverInterface
     public function resolve()
     {
         if (!isset($_SERVER['REQUEST_METHOD'])) {
-            return new NullRequest();
```
GrahamCampbell: This method could be updated to avoid the nested ifs:

```php
    /**
     * Resolve the current request.
     *
     * @return \Bugsnag\Request\RequestInterface
     */
    public function resolve()
    {
        if (isset($_SERVER['REQUEST_METHOD'])) {
            return new PhpRequest(
                $_SERVER,
                empty($_SESSION) ? [] : $_SESSION,
                empty($_COOKIE) ? [] : $_COOKIE,
                static::getRequestHeaders($_SERVER),
                static::getInputParams($_SERVER, $_POST)
            );
        }

        if (PHP_SAPI === 'cli' && isset($_SERVER['argv'])) {
            return new ConsoleRequest($_SERVER['argv']);
        }

        return new NullRequest();
    }
```

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Get the request context.
+     *
+     * @return string|null
+     */
+    public function getContext()
+    {
+        return implode(' ', array_slice($this->command, 0, 2));
```
GrahamCampbell: Just realized that php doesn't include the executable in the arguments list. Maybe we should just take the first part of the array then, or implode the whole thing. Ping @kattrali - what do you think is the most sensible option here?

kattrali: I think first few args still works well enough. The cases I can try/think of:
* `php [filename]` => `[filename]`
* `php [filename] foo bar` => `[filename] foo`
* `php [filename] --foo bar --baz red` => `[filename] --foo` <-- this one is a lil odd
* `php artisan migrate` => `artisan migrate`

I could see the value in imploding the whole thing, but also don't want context to get arbitrarily long. We could max four args for now and revisit with feedback.

GrahamCampbell: > but also don't want context to get arbitrarily long

Yeh, totally. We need to find the balance where it is useful but not unbounded.

src/Request/ConsoleRequest.php
kattrali: Haha, explains the deja vu I felt when I typed that :p

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,102 @@
...
+     * @return array
+     */
+    public function getMetaData()
+    {
+        $commandString = implode(' ', $this->command);
+        $primaryCommand = $this->command[0];
+        $arguments = [];
+        $options = [];
+        foreach (array_slice($this->command, 1) as $arg) {
+            if ($arg[0] == '-') {
```
GrahamCampbell: Safest option is to do: `if (isset($arg[0]) && $arg[0] === '-') {`.

tests/Callbacks/RequestMetaDataTest.php
```diff
@@ -49,8 +49,28 @@ public function testCanMetaData()
...
+        $callback = new RequestMetaData($this->resolver);
+
+        $callback($report);
+
+        $this->assertSame(['bar' => 'baz', 'console' => [
+            'Input' => 'some test command --opt',
+            'Command' => 'some',
+            'Arguments' => ['test', 'command'],
+            'Options' => ['--opt'],
+        ]], $report->getMetaData());
```
GrahamCampbell: Probably worth also doing an assertion on the report context too.

GrahamCampbell:
Left 2 more tiny comments. Otherwise, looks good. 🚀 🚀 🚀

tests/Callbacks/RequestMetaDataTest.php
kattrali: @GrahamCampbell Should there be a test which is different from [this one](https://github.com/bugsnag/bugsnag-php/pull/398/files#diff-f9d661e66db5a9fa0ced0d87ae7ed85fR52)?

GrahamCampbell: Oh, ha. No, that looks great. ;)

kattrali:
🚢 🚢 🚢

PR #406, opened 2017-09-13
Adds Handled unhandled properties to payload

Cawllec:
- Adds `defaultSeverity`, `unhandled`, and `severityReasons` payload properties to support upcoming handled/unhandled functionality
- Properties added when creating a report, functionality a user should not use directly
- Systems error handlers setup with correct severityReasons
_references [PLAT_208](https://bugsnag.atlassian.net/browse/PLAT-208)_

### DIFF
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -136,7 +136,7 @@ class Handler
      */
     public function exceptionHandler($throwable)
     {
-        $report = Report::fromPHPThrowable($this->client->getConfig(), $throwable);
+        $report = Report::fromPHPThrowable($this->client->getConfig(), $throwable, Report::EXCEPTION_HANDLER);
 
         $report->setSeverity('error');
 
@@ -163,7 +163,7 @@ class Handler
     public function errorHandler($errno, $errstr, $errfile = '', $errline = 0)
     {
         if (!$this->client->shouldIgnoreErrorCode($errno)) {
-            $report = Report::fromPHPError($this->client->getConfig(), $errno, $errstr, $errfile, $errline);
+            $report = Report::fromPHPError($this->client->getConfig(), $errno, $errstr, $errfile, $errline, false, Report::EXCEPTION_HANDLER);
 
             $this->client->notify($report);
         }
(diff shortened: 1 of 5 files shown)


src/Handler.php
```diff
@@ -163,7 +163,7 @@ public function exceptionHandler($throwable)
     public function errorHandler($errno, $errstr, $errfile = '', $errline = 0)
     {
         if (!$this->client->shouldIgnoreErrorCode($errno)) {
-            $report = Report::fromPHPError($this->client->getConfig(), $errno, $errstr, $errfile, $errline);
+            $report = Report::fromPHPError($this->client->getConfig(), $errno, $errstr, $errfile, $errline, false, Report::EXCEPTION_HANDLER);
```
snmaynard: This severity reason should probably be ERROR_CLASS because we set the severity based on the error raised

tests/ReportTest.php
```diff
@@ -298,4 +298,32 @@ public function testGetSummaryDuplicate()
 
         $this->assertSame(['message' => 'bar', 'severity' => 'warning'], $this->report->getSummary());
     }
+
+    /**
+     * Testing handled/unhandled.
+     */
+    public function testDefaultsToHandled()
```
snmaynard: Can we add defaultseverity tests? Do we have a ticket to add code coverage stats to the tests?

Cawllec:
I've added tests for defaultSeverity and updated the error handler severityReason

tests/ReportTest.php
Cawllec: I've added some default Severity tests in

GrahamCampbell:
Technically, adding new parameters to non-private methods on non-final classes is a major breaking change.

GrahamCampbell:
This can be mitigated by just calling the setter methods on the report where needed.

Cawllec:
I'll revisit this and see if it can be changed to use the setter methods.  Now that we're preserving `severity_reason` before callbacks/middleware it should be less of an issue

PR #410, opened 2017-09-29
feature: allow to limit notified hostnames via config

janmyszkier:


snmaynard:
Thanks for the contribution @janmyszkier! Can you use a callback that you return false from to achieve this? https://docs.bugsnag.com/platforms/php/other/customizing-error-reports/#adding-callbacks for docs on that

janmyszkier:
Hey @snmaynard I don't think I can get this done by just metadata. I see no way to disable email/slack notifications by JUST metadata. or is there just UI config I'm missing?

Cawllec:
Hey @janmyszkier as an example of using callbacks to block specific hosts from reporting you can implement something like -
```ruby
$bugsnag->registerCallback(function ($report) {
    $allowedHostnames = ["hostname_A", "hostname_B"];
    $hostname = gethostname();
    if (!in_array($hostname, $allowedHostnames, false)) {
        return false;
    }
});
```
If you want more information specifically for Symfony check out [customizing error reports](https://docs.bugsnag.com/platforms/php/symfony/customizing-error-reports/).

I think callbacks are the preferred method for this sort of functionality, it avoids us getting overrun with configuration options :).

janmyszkier:
@Cawllec I don't think you read the case we're having:
- let's say you have 3 different environments (demo / staging / live)
- each of them has bugsnag-symfony as a dependency (in vendor, no sane people modify vendor )

Considering I'm letting bugsnag report unhandled exceptions just by adding API KEY ( explained here: https://docs.bugsnag.com/platforms/php/symfony/#basic-configuration ) 

how do I stop it from reporting non-dev/non-local errors that the other way than:
a) create my own bundle 
b) use my own modified-fork of bugsnag-symfony with hostnames filter
?

the funtion you're pointing at is within my vendor folder right now

janmyszkier:
@Cawllec :point_up: your `user feedback`. the label is not correct anymore

Cawllec:
Ok, in trying to produce an example of what you want I can see that there isn't an easy place to put callbacks or to set up the library.  
My first thought was to use a `ContainerExtension` to register the callbacks except the `client` can't be retrieved at boot due to using some security objects for user identification.
As an alternative to using hostnames directly, can you use `release_stage` and `notify_release_stages ` to get the behaviour you require, as described [here](https://docs.bugsnag.com/platforms/php/symfony/configuration-options/#release-stage)?

Issue #411, opened 2017-10-05
Laravel Lumen does not receive server argv in LaravelResolver, throws uncaught exception


georgeboot:
### Expected behavior
The app successfully reports the exception to Bugsnag

### Observed behavior
The Bugsnag code throws an exception: `PHP Fatal error:  Uncaught ErrorException: Undefined offset: 0 in /path/to/project/vendor/bugsnag/bugsnag/src/Request/ConsoleRequest.php:64`

### Steps to reproduce

1. Install a fresh Lumen 5.5 application
2. Follow the instructions [here](https://docs.bugsnag.com/platforms/php/lumen/) to install bugsnag into the lumen project
3. Trigger an uncaught exception within the console.
4. The above mentioned exception will be thrown

### Version

- Lumen 5.5.0
- bugsnag/bugsnag v3.8.0
- bugsnag/bugsnag-laravel v2.9.0

Cawllec:
Hi @georgeboot thanks for letting us know.  
I've written a fix for this that'll use a sensible default instead of throwing an exception #413 . 
I'll look into getting a further fix written to ensure that we can correctly extract console arguments from Lumen and report them.

georgeboot:
Hi @Cawllec,
I can conform that your fix on master fixed the issue. Thanks for the quick support!

PR #413, opened 2017-10-05
Ensured can handle empty console arg array correctly

Cawllec:
Fixes issue reported here #411 

- Ensures defaults in place if an empty array is provided to as console commands.


src/Request/ConsoleRequest.php
```diff
@@ -60,6 +60,11 @@ public function getCookies()
      */
     public function getMetaData()
     {
+        if (count($this->command) == 0) {
+            return ['console' => [
+                'Command' => 'Command could not be retreived', ],
```
kattrali: `retrieved`

PR #419, opened 2017-11-01
Adds Bugsnag-logger configuration options

Cawllec:
This adds 3 new configuration options for use with the [Bugsnag PSR logger](https://github.com/bugsnag/bugsnag-psr-logger):

NotifyLevel: The LogLevel at which notifications will be sent to bugsnag
WarningLevel: The LogLevel at which notifications will be given a severity "warning"
ErrorLevel: The LogLevel at which notifications will be given a severity "error"

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -102,6 +102,27 @@ class Configuration
     protected $errorReportingLevel;
 
     /**
+     * The level logs must match or exceed to send a notification
+     * 
+     * @var string|null
+     */
+    protected $logNotifyLevel;
+
+    /**
+     * The level logs must match or exceed be set to a "warning" severity
+     * 
+     * @var string|null
+     */
+    protected $logWarningLevel;
+
+    /**
+     * The level logs must match or exceed be set to an "error" severity
+     * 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Is this not overconfiguration? Why would anybody want to change this?

Cawllec:
At the moment log-levels aren't configurable at all in the PSR logger, which really should be standard behaviour.  I'm not sure if we want to keep the Warning & Error level config options however, but we should have a notify level setting. 
I was going to add extra config in the constructor of the logger (to be less obtrusive) but that wouldn't have helped in cases like laravel where a class is being aliased instead of constructed.

snmaynard:
I like the notify level config. If the severities map fairly sensibly to log levels then I wouldnt add config for it personally.

Cawllec:
I've removed the config options for warning/error levels, and updated [#23](https://github.com/bugsnag/bugsnag-psr-logger/pull/23) in the PSR logger appropriately

Cawllec:
Closing this in favour of logger construction arguments instead

PR #421, opened 2017-11-01
Fix: Ensure that the "type" parameter has a sensible default

Cawllec:
Ensures that the "type" parameter in the "severityReason" has a sensible default.


GrahamCampbell:
What is wrong with the way it is right now?

PR #423, opened 2017-11-11
Added CodeIgniter example

donpinkus:
Added a readme and example project to /example/code-igniter


GrahamCampbell:
Wow, that's a lot of files. Perhaps this belongs in another tutorial repo?

kattrali:
Wow, I've got a couple of questions, @donpinkus:

1. How do you run this?
2. Can you use something like the [composer installer](https://github.com/kenjis/codeigniter-composer-installer) to stick to a composer-only development flow? Between that and removing the `user_guide` files, this would become a much more manageable example, hopefully from 200k lines of code down to maybe ~1k.

donpinkus:
@GrahamCampbell - probably, I'm fine moving it to a separate repo and just linking to it from the README here and the pages on our site that reference it.

@kattrali:
1. How do you run this?
I gave pretty detailed instructions in the [README](https://github.com/donpinkus/bugsnag-php/blob/db7affe3a900bdc02d1700985e438a0e066b51af/example/code-igniter/README.md), was there some step in particular that's unclear?

PR #450, opened 2018-01-03
Session tracking

Cawllec:
Implementation of session tracking.  Adds the ability to automatically or manually track request sessions.


composer.json
```diff
@@ -16,6 +16,8 @@
         "guzzlehttp/guzzle": "^5.0|^6.0"
     },
     "require-dev": {
+        "graham-campbell/testbench-core": "^1.1",
+        "mockery/mockery": "^0.9.4",
```
GrahamCampbell: Mockery 1.x is newest.

src/Configuration.php
```diff
@@ -486,4 +507,48 @@ public function shouldIgnoreErrorCode($code)
...
+        if (!$track) {
+            $this->trackSessions = false;
+
+            return;
+        }
+        if (is_null($endpoint)) {
+            $endpoint = self::SESSION_ENDPOINT;
+        }
+        $this->trackSessions = true;
+        $this->sessionClient = Client::makeGuzzle($endpoint);
```
GrahamCampbell: Hmm, not sure I like the config class being able to do this.

src/Middleware/SessionData.php
```diff
@@ -0,0 +1,51 @@
...
+     * The client instance.
+     *
+     * @var \Bugsnag\Client
+     */
+    protected $client;
+
+    /**
+     * Create a new session data middleware instance.
+     *
+     * @param \Bugsnag\SessionTracker $client the client instance.
```
GrahamCampbell: This type doesn't match the type in the code. :)

src/SessionTracker.php
```diff
@@ -0,0 +1,326 @@
...
+        $this->config = $config;
+    }
+
+    public function createSession()
+    {
+        if (!$this->config->shouldTrackSessions()) {
+            return;
+        }
+        $currentTime = strftime('%Y-%m-%dT%H:%M:00');
+        $session = [
```
GrahamCampbell: Might be worth having a session object that has a `toArray` method.

src/Configuration.php
```diff
@@ -486,4 +507,48 @@ public function shouldIgnoreErrorCode($code)
 
         return !(error_reporting() & $code);
     }
+
+    /**
+     * Set session tracking state and pass in optional guzzle.
+     *
+     * @param bool $track whether to track sessions
+     * @param string $endpoint an optional session endpoint
```
GrahamCampbell: `string|null` is the type here :)

src/SessionTracker.php
```diff
@@ -0,0 +1,326 @@
...
+     * An array of session counts.
+     *
+     * @var array
+     */
+    protected $sessionCounts = [];
+
+    /**
+     * A locking function for synchronisation.
+     *
+     * @var function
```
GrahamCampbell: this is not a valid type in php, you perhaps want `callable`?

GrahamCampbell: Though, this type is actually `callable|null`

GrahamCampbell: This `callable` type exists in php 5.4+, and is a pesudo-type in php 5.3.

src/SessionTracker.php
```diff
@@ -0,0 +1,326 @@
...
+            return call_user_func($this->storageFunction, 'bugsnag-session-counts');
+        } else {
+            return $this->sessionCounts;
+        }
+    }
+
+    protected function setSessionCounts(array $sessionCounts)
+    {
+        if (!is_null($this->storageFunction)) {
+            return call_user_func($this->storageFunction, 'bugsnag-session-counts', $sessionCounts);
```
GrahamCampbell: What is this `'bugsnag-session-counts'` string?

GrahamCampbell: Oh, I see. Maybe make it a constant in this class?

src/SessionTracker.php
```diff
@@ -0,0 +1,326 @@
...
+            $this->setSessionCounts($sessionCounts);
+
+            if (count($sessionCounts) > self::$MAX_SESSION_COUNT) {
+                $this->trimOldestSessions();
+            }
+            $lastSent = $this->getLastSent();
+            if ($deliver && ((time() - $lastSent) > self::$DELIVERY_INTERVAL)) {
+                $this->deliverSessions();
+            }
+        } finally {
```
GrahamCampbell: The finally keyword is a php 5.5+ feature, so won't work on php 5.3.

src/SessionTracker.php
```diff
@@ -0,0 +1,326 @@
...
+    {
+        $locked = false;
+        if (!is_null($this->lockFunction) && !is_null($this->unlockFunction)) {
+            call_user_func($this->lockFunction);
+            $locked = true;
+        }
+
+        try {
+            $this->deliverSessions();
+        } finally {
```
GrahamCampbell: Another use of php 5.5+ features here.

composer.json
GrahamCampbell: `^1.0`

GrahamCampbell: `^1.0`

src/SessionTracker.php
Cawllec: As I understand our minimum supported version is 5.5.9, so this shouldn't be an issue?

src/Configuration.php
Cawllec: I agree.  Ideally it would have been better to somehow merge this functionality into the HttpClient, which I may still do.

src/SessionTracker.php
GrahamCampbell: Oh, yeh, the minimum is 5.5.0 in this package. Looking at the history It was bumped from 5.2.0 to 5.5.9 in the v3.0.0 release, then moved down to 5.5.0 in v3.0.1+.

GrahamCampbell: How do you dismiss reviews in GitHub? :P

composer.json
```diff
@@ -17,7 +17,7 @@
     },
     "require-dev": {
         "graham-campbell/testbench-core": "^1.1",
-        "mockery/mockery": "^1.0",
+        "mockery/mockery": "^0.9.4|^1.0",
```
GrahamCampbell: You shouldn't need to add the old version?

src/Configuration.php
```diff
@@ -486,4 +514,61 @@ public function shouldIgnoreErrorCode($code)
...
+    }
+
+    /**
+     * Get the session client.
+     *
+     * @return \Guzzle\ClientInterface
+     */
+    public function getSessionClient()
+    {
+        if (!$this->autoCaptureSessions) {
```
kattrali: It should be possible to get a client event when auto capture is disabled

IMM0rtalis:
@Cawllec @kattrali I am really wondering about where `shouldCaptureSessions()` is used. As far as I can tell this configuration value has not any effect. 

Background: I wanted to switch off the session sending but this function seems like doing nothing and is not used at all.

PR #454, opened 2018-01-10
Updates deploy to build API

Cawllec:
Updates to the new Bugsnag build API.
- Adds `build` method to client to send build information off to bugsnag
- Points clients `deploy` method to `build` method
- Maintains old `deploy` method in HTTP client unchanged
- Adds new `sendBuildReport` to HTTP client to handle new API calls.
- Adds `setBuildEndpoint` and `getBuildEndpoint` to configuration, with a default.

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -134,6 +134,12 @@ class HttpClient
             $data['builderName'] = $buildInfo['builder'];
         }
 
+        if (isset($buildInfo['buildTool'])) {
+            $data['buildTool'] = $buildInfo['buildTool'];
+        } else {
+            $data['buildTool'] = 'bugsnag-php';
+        }
+
         $data['releaseStage'] = $app['releaseStage'];
 
         if (isset($app['version'])) {
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -124,7 +124,7 @@ class HttpClient
 
         if (isset($buildInfo['builder'])) {
             $data['builderName'] = $buildInfo['builder'];
-        } else {
+        } elseif ($this->functionAvailable('exec')) {
             $data['builderName'] = exec('whoami');
         }
 
@@ -145,6 +145,19 @@ class HttpClient
         $endpoint = $this->config->getBuildEndpoint();
 
         $this->guzzle->post($endpoint, ['json' => $data]);
+    }
+
+    /**
+     * Checks whether the function is available
+     *
+     * @param string $func function name
+     *
+     * @return bool
+     */
+    protected function functionAvailable($func)
+    {
+        $disabled = explode(',', ini_get('disable_functions'));
+        return function_exists($func) && !in_array($func, $disabled);
     }
 
     /**
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -3,6 +3,7 @@
 namespace Bugsnag;
 
 use Exception;
+use Bugsnag\Utils;
 use GuzzleHttp\ClientInterface;
 use RuntimeException;
 
@@ -124,8 +125,8 @@ class HttpClient
 
         if (isset($buildInfo['builder'])) {
             $data['builderName'] = $buildInfo['builder'];
-        } elseif ($this->functionAvailable('exec')) {
-            $data['builderName'] = exec('whoami');
+        } else{
+            $data['builderName'] = Utils::getBuilderName();
         }
 
         if (isset($buildInfo['buildTool'])) {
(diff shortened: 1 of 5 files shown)


src/HttpClient.php
```diff
@@ -91,6 +93,67 @@ public function deploy(array $data)
...
+        $sourceControl = [];
+        if (isset($buildInfo['repository'])) {
+            $sourceControl['repository'] = $buildInfo['repository'];
+        }
+
+        if (isset($buildInfo['provider'])) {
+            $sourceControl['provider'] = $buildInfo['provider'];
+        } elseif (isset($buildInfo['repository'])) {
+            $url = $buildInfo['repository'];
+            if (strpos($url, 'github.com') != false) {
```
snmaynard: You dont need to do this. Provider is optional

src/HttpClient.php
```diff
@@ -91,6 +93,67 @@ public function deploy(array $data)
...
+        if (!empty($sourceControl)) {
+            $data['sourceControl'] = $sourceControl;
+        }
+
+        if (isset($buildInfo['builder'])) {
+            $data['builderName'] = $buildInfo['builder'];
+        }
+
+        if (isset($buildInfo['buildTool'])) {
+            $data['buildTool'] = $buildInfo['buildTool'];
```
snmaynard: You shouldnt be able to externally set this

src/HttpClient.php
```diff
@@ -91,6 +93,67 @@ public function deploy(array $data)
...
+        if (isset($buildInfo['revision'])) {
+            $sourceControl['revision'] = $buildInfo['revision'];
+        }
+
+        if (!empty($sourceControl)) {
+            $data['sourceControl'] = $sourceControl;
+        }
+
+        if (isset($buildInfo['builder'])) {
+            $data['builderName'] = $buildInfo['builder'];
```
snmaynard: This should default to `whoami` or a php cross platform equivalent.

src/HttpClient.php
kattrali: Its settable for laravel to override

src/HttpClient.php
Cawllec: I don't think we want to add shell calls in the base notifier, it's much safer to use wrapped commands in higher level libs.

src/HttpClient.php
snmaynard: I dont think we care about php/laravel distinction at an analytics level. We just care that this lib is being used for deploys.

src/HttpClient.php
snmaynard: Why not? If this isnt automated very few people will set it.

src/HttpClient.php
kattrali: Every field is optional though. Are we using this field for something?

snmaynard: We shouldnt automatically set this in any of our sdks. This logic of guessing the provider is on the server (and therefore is easier to update).

The user needs to be able to set it for when they are using an on-premise installation of github and the repository URL isnt the github.com one.

Cawllec: It gives us the extra data they may not set otherwise.

Cawllec: Ah, ok, didn't see this.

src/HttpClient.php
```diff
@@ -91,6 +93,60 @@ public function deploy(array $data)
...
+        } else {
+            $data['builderName'] = exec('whoami');
```
GrahamCampbell: You should check if this function has been disabled, otherwise it'll crash. Many web hosts disable this by default for security.

src/HttpClient.php
GrahamCampbell: You probably want:

```php
get_current_user();
```

instead of shell executing `whoami`. That function will give you the owner of the file, which is maybe better than nothing.

Something you could do if the user has installed and not disabled the functions from posix extension, is:

```php
posix_getpwuid(posix_geteuid())['name'];
```

Cawllec: `whoami` works across all windows post server-2003 & unix-like systems.  I'll add a check on the response to see if it returned correctly and fallback to `get_current_user` if not.

jbrooksuk:
I was just looking for this and was about to contribute it. Is there any chance we can see this get merged and released soon?

Cawllec:
Yeah, i'll get this released this afternoon if possible.

jbrooksuk:
Awesome, thanks @Cawllec!

PR #461, opened 2018-01-23
Reject empty build version

Cawllec:
Adds an error message and halts the sending of a build report if the `appVersion` is not set.

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -106,6 +106,14 @@ class HttpClient
 
         $data = [];
         $sourceControl = [];
+
+        if (isset($app['version'])) {
+            $data['appVersion'] = $app['version'];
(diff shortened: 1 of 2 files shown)


cmorbitzer:
Hey, it's kind of irritating that this breaking change was introduced without documentation. Prior to this change, releases would be marked by timestamp which works well for us because we develop a hosted SaaS product with continuous updates. No need to label releases.

Unfortunately, the release script documented here: https://docs.bugsnag.com/platforms/php/laravel/#tracking-releases (which states we could send a "source revision **or** application version", we always send source revision) started to fail silently for us after the release of this PR. As a result, Bugsnag never tracked new releases for us and thought that new bugs were duplicates of old bugs and we missed them until just now.

PR #467, opened 2018-02-22
Updates the payload body `payloadVersion`

Cawllec:
Ensures the `PAYLOAD_VERSION` static in `Report.php` is consistent with `PAYLOAD_VERSION` in `HttpClient.php`.  
Added tests to check the payload version is set correctly in both headers and body.

### DIFF
--- a/src/Report.php
+++ b/src/Report.php
@@ -9,12 +9,13 @@ use Throwable;
 
 class Report
 {
+
     /**
      * The payload version.
      *
      * @var string
      */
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Kinda odd that the http client needs to know the payload version?

Cawllec:
Payload version now effects the body and the headers of the request, so HttpClient makes more sense to me as it'll be dealing with both.

PR #470, opened 2018-03-16
Ensure name represents class before reflecting

funkjedi:
Test that the breadcrumb name represents a valid class before reflecting. Without this debugging can be a nightmare due to all the unnecessary ReflectionExceptions being thrown.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -211,10 +211,12 @@ class Client
      */
     public function leaveBreadcrumb($name, $type = null, array $metaData = [])
     {
-        try {
-            $name = (new ReflectionClass($name))->getShortName();
-        } catch (ReflectionException $e) {
-            //
+        if (class_exists($name)) {
+            try {
+                $name = (new ReflectionClass($name))->getShortName();
+            } catch (ReflectionException $e) {
+                //
+            }
         }
 
         $name = substr((string) $name, 0, Breadcrumb::MAX_LENGTH);


GrahamCampbell:
This just duplicates the logic through. Now PHP has to run the same code twice.

funkjedi:
If duplication is a concern how would you feel about skipping the reflection all together? Just check if the class exists and strip the namespace if there's a match.

```php
if (class_exists($name)) {
	$name = ltrim(strrchr($name, '\\') ?: $name, '\\');
}
```

GrahamCampbell:
TBH, I don't think there's any reason to change the current implementation. It's reusing what the language makes available to us. I think it's fine.

funkjedi:
Xdebug breaks on all caught exceptions. As a consequence, under the current implementation, adding a breadcrumb also adds a breakpoint. 

Even if you only have a dozen or so breadcrumbs it's enough make debugging unnecessarily tedious.

If the aforementioned proposals aren't acceptable what about adding a `getRecorder()` method so breadcrumbs can be added directly.

```php
public function getRecorder()
{
    return $this->recorder;
}
```

johnkiely1:
This has now been fixed via release V3.21.0
https://github.com/bugsnag/bugsnag-php/releases/tag/v3.21.0.

The reflection has been removed from the code

PR #472, opened 2018-03-26
feature(resolver): Add support for get parameters

Cawllec:
- Adds support for `get` parameters in the basic resolver
- Retains support for `$_POST` and other method parameters.

### DIFF
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -17,6 +17,7 @@ class BasicResolver implements ResolverInterface
             } else {
                 $params = static::getInputParams($_SERVER, $_POST, false);
             }
+
             return new PhpRequest($_SERVER,
                 empty($_SESSION) ? [] : $_SESSION,
                 empty($_COOKIE) ? [] : $_COOKIE,
@@ -80,7 +81,7 @@ class BasicResolver implements ResolverInterface
      *
      * @return array|null
      */
-    protected static function getInputParams(array $server, array $params, bool $is_get=false)
+    protected static function getInputParams(array $server, array $params, bool $is_get = false)
     {
         static $result;
 
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -119,6 +119,7 @@ class BasicResolver implements ResolverInterface
             return (array) json_decode($input, true) ?: null;
         } else {
             parse_str($input, $params);
+
             return (array) $params ?: null;
         }
     }
--- a/src/Request/BasicResolver.php
+++ b/src/Request/BasicResolver.php
@@ -78,10 +78,11 @@ class BasicResolver implements ResolverInterface
      *
      * @param array $server the server variables
      * @param array $params the array of parameters for this request type
+     * @param bool  $isGet  whether this is a get request
      *
      * @return array|null
      */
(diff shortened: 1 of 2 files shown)


tests/RequestTest.php
```diff
@@ -103,6 +103,23 @@ public function testGetMetaDataWithPost()
         $this->assertSame(['request' => $data], $this->resolver->resolve()->getMetaData());
     }
 
+    public function testGetMetaDataWithGet()
+    {
+        $_SERVER['REQUEST_METHOD'] = 'GET';
```
fractalwrench: What would happen with lesser used request methods like `HEAD` or `PATCH`? Is it worth covering them?

Cawllec: They don't have server variables and the data should be dropped into either the `json_decode` section due to `application/json` being present, or the `parse_str` section if not.  I'll do some manual testing on this to verify these cases are covered however.

src/Request/BasicResolver.php
```diff
@@ -71,19 +71,22 @@ protected static function getRequestHeaders(array $server)
...
     {
         static $result;
 
         if ($result !== null) {
             return $result ?: null;
         }
 
-        $result = $post ?: static::parseInput($server, static::readInput());
+        if (isset($server['REQUEST_METHOD']) && strtoupper($server['REQUEST_METHOD']) === 'GET') {
+            $result = $_GET;
```
GrahamCampbell: Access to superglobals was intentionally omitted from anything other than the resolve method. Perhaps put it back there, so that this method is not messing with global state?

Cawllec: Does it matter if the method is only being called statically from within the `resolve` method?

src/Request/BasicResolver.php
```diff
@@ -16,7 +16,7 @@ public function resolve()
                 empty($_SESSION) ? [] : $_SESSION,
                 empty($_COOKIE) ? [] : $_COOKIE,
                 static::getRequestHeaders($_SERVER),
-                static::getInputParams($_SERVER, $_POST));
+                static::getInputParams($_SERVER));
```
kattrali: Checking the request method should instead happen here, and pass the correct variable array (get vs post variables) into `getInputParams`. It reduces the amount of "magic" going on throughout.

Cawllec: In the current implementation if `$_GET` is empty in a `GET` request the method will attempt to pull body parameters from `php://input`, something that it shouldn't do for `GET` requests. 
I'd suggest splitting `getInputParams` into multiple methods to handle `GET` and other request methods separately, or add an argument that determines whether the `php://input` source should be read or not.

kattrali: Correct, checking whether to read from `php://input` should be done by request type, either in a single place or via a flag. I'm not suggesting that input should be read for GET requests.

Cawllec:
Ok, I've refactored it somewhat so that the superglobals `$_SERVER`, `$_POST`, and `$_GET` are accessed only in the `resolve` method, and passed through to `getInputParams` as before.

I've added a flag, `$is_get` to the method in order to determine whether the `parseInput` methods should be called.

The result doesn't effect the output, so the tests have remained the same.

PR #476, opened 2018-05-19
Speed up error handler

snmaynard:
The reflection we are using here is reducing performance. We should also long term remove the proxy method we have from Client to Config as it adds an unacceptable performance overhead.

This is untested.

### DIFF
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -169,7 +169,7 @@ class Handler
      */
     public function errorHandler($errno, $errstr, $errfile = '', $errline = 0)
     {
-        if (!$this->client->shouldIgnoreErrorCode($errno)) {
+        if (!$this->client->getConfig()->shouldIgnoreErrorCode($errno)) {
             $report = Report::fromPHPError(
                 $this->client->getConfig(),
                 $errno,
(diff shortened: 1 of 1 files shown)


loopj:
LGTM, agree on removing reflection completely too.

thedotedge:
@GrahamCampbell to add more context, performance hit can be quite significant for some scenarios. This is a trace from php 7.1 where you can see severe impact:
image

GrahamCampbell:
@thedotedge I find that Backfire tends to exaggerate the cost of lots of small functions calls, compared to the actual runtime when the profiler is turned off. In particular, I found this when profiling the Laravel container, where Blackfire said it was using half the CPU time. How significant does this compare to other code. It is called 7900 times, so it's taking way less than 0.01ms per time?

snmaynard:
In that screenshot its not showing the Bugsnag\Client::shouldIgnoreErrorCode call which is the bit that this would remove IIRC

thedotedge:
@snmaynard yep, but you have a full trace :)

thedotedge:
Guys, 10 days after PR was open and still no release? 😢

thedotedge:
ping :)

PR #478, opened 2018-05-21
feature(pipeline): add support for registering middleware

Cawllec:
## Goal

This change adds a `registerMiddleware` method to the  `Client` class that allows middleware to be directly added to the pipeline.  It also shifts the consistency checks on `unhandled` and `severity` to the `CallbackBridge` middleware, allowing registered middleware to modify these elements.

## Design

Interface changes:
`public function registerMiddleware(callable $middleware)`

Logic changes:
Wrapped given callback in `notify` in a `CallbackBridge`, complicating logic around calling it.  Any feedback on that would be appreciated.

### Added
`registerMiddleware` method

### Changed
Moved severity consistency checks to the `CallbackBridge` middleware

## Tests

Added unit tests

## Discussion

I'm not especially happy with the callback logic changed in `notify`, so any suggestions on how to neaten/improve on that would be appreciated.


src/Client.php
```diff
@@ -200,6 +200,24 @@ public function registerDefaultCallbacks()
...
+    /**
+     * Register a middleware object to the pipeline.
+     *
+     * @param callable $middleware
+     *
+     * @return $this
+     */
+    public function registerMiddleware(callable $middleware)
+    {
+        if (is_callable($middleware)) {
```
GrahamCampbell: Don't need to check this. The type annotation already raises an error.

src/Client.php
```diff
@@ -200,6 +200,24 @@ public function registerDefaultCallbacks()
...
+            $this->pipeline->pipe($middleware);
+        } else {
+            syslog(LOG_WARNING, 'Middleware '.get_class($middleware).' could not be added to the pipeline');
```
GrahamCampbell: This code is not reachable, and should be deleted.

Cawllec: You're correct, I was just verifying the earlier versions of PHP had the same behaviour

GrahamCampbell: They should do. Callable was introduced in php 5.4 as a typehint, and the behavior has remained the same since.

Cawllec: I'm not sure what kind of error is thrown in earlier versions of PHP, in PHP 7 it's quite explicitly a TypeError, any idea for PHP 5.6?

GrahamCampbell: https://3v4l.org/FnNTN

PR #484, opened 2018-07-05
Next release

Cawllec:
PR for 3.14.0 release


GrahamCampbell:
The branch alias in the composer.json also needs updating.

NB Using a next branch instead of committing dev work to master like was previously done has the disadvantage that composer can't fetch dev versions of the software.

PR #492, opened 2018-10-03
Fixed recursion on payload too large

woutersamaey:
## Goal

Fixed the endless recursion when the payload to submit is too large. The PHP max function nesting will be reached.

## Review

  - [x] Structural review of the classes, functions, and properties modified
  - [x] Final review

### DIFF
--- a/src/Exception/PayloadTooLargeException.php
+++ b/src/Exception/PayloadTooLargeException.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Bugsnag\Exception;
+
+class PayloadTooLargeException extends \RuntimeException {
+
+    private $payload;
+
+    public function __construct(array $payload) {
+        parent::__construct('Payload too large', 0, null);
+
+        $this->payload = $payload;
+    }
+
+    /**
+     * @return array
+     */
+    public function getPayload() {
+        return $this->payload;
+    }
+
+
+}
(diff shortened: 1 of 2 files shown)


woutersamaey:
We had an exception with a huge message that could not be sent to Bugsnag.
This made debugging the issue on production very hard.

The idea is that if bugsnag can't take the exception, then at least it should help me out instead of failing with another error, masking the original.

The underlying issue was easy to fix for me, but because bugsnag could not handle it, it became a huge thing. So that's why I wrote the message out.

Cawllec:
I agree, that doesn't sounds ideal at all.  I think adding a failure condition with full logs would be a practical way to approach this as suggested, and I'll suggest we prioritise this to get a fix as soon as we can.

GrahamCampbell:
> What about creating a fallback LoggerInterface where we can attach our own logic to, so that will be used in case Bugsnag can't handle it?

For now, I think it's best to do what the library already does, which is to fallback to php's own global logging functions. Changing how the library deals with internal exceptions is a separate discussion to be had I think, if we want to change that. One of the reasons for the current design is to prevent the possibility non-termination due to bugsnag being the psr logger itself.

GrahamCampbell:
> For now, I think it's best to do what the library already does

Ah, looking at your changes, I see that's what you've done. Perfect! :)

woutersamaey:
Huh?

PR #495, opened 2018-10-24
Cawllec/before class middleware

Cawllec:
## Goal
Allow middleware to be inserted before specific classes in the `$pipeline`.

## Design
- See design doc.

### Added
- `stringEndsIn` static method in Utils module.
- `insertBefore` method in `Pipeline` class.

### Changed
- `$pipeline` instance is now public in `Client` class.

## Tests
Added basic `Pipeline` tests, alongside `insertBefore` tests.

## Questions
- What test cases are required?
- What docs changes are necessary?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -58,7 +58,7 @@ class Client
      *
      * @var \Bugsnag\Pipeline
      */
-    protected $pipeline;
+    public $pipeline;
 
     /**
      * The http client instance.
--- a/src/Client.php
+++ b/src/Client.php
@@ -117,9 +117,9 @@ class Client
         $this->http = new HttpClient($config, $guzzle ?: static::makeGuzzle());
         $this->sessionTracker = new SessionTracker($config);
 
-        $this->pipeline->pipe(new NotificationSkipper($config));
-        $this->pipeline->pipe(new BreadcrumbData($this->recorder));
-        $this->pipeline->pipe(new SessionData($this));
+        $this->registerMiddleware(new NotificationSkipper($config));
+        $this->registerMiddleware(new BreadcrumbData($this->recorder));
+        $this->registerMiddleware(new SessionData($this));
 
         register_shutdown_function([$this, 'flush']);
     }
@@ -178,7 +178,7 @@ class Client
      */
     public function registerCallback(callable $callback)
     {
-        $this->pipeline->pipe(new CallbackBridge($callback));
+        $this->registerMiddleware(new CallbackBridge($callback));
 
         return $this;
     }
@@ -196,6 +196,20 @@ class Client
              ->registerCallback(new RequestSession($this->resolver))
              ->registerCallback(new RequestUser($this->resolver))
              ->registerCallback(new RequestContext($this->resolver));
+
+        return $this;
+    }
+
+    /**
+     * Register a middleware object to the pipeline.
+     *
+     * @param callable $middleware
+     *
+     * @return $this
+     */
+    public function registerMiddleware(callable $middleware)
+    {
+        $this->pipeline->pipe($middleware);
 
         return $this;
     }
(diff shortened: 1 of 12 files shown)


src/Pipeline.php
```diff
@@ -37,6 +37,33 @@ public function pipe(callable $pipe)
...
+     * @param string $beforeClass to class to insert the pipe before
+     *
+     * @return $this
+     */
+    public function insertBefore(callable $pipe, $beforeClass)
+    {
+        $beforePosition = null;
+        foreach ($this->pipes as $index => $callable) {
+            $class = get_class($callable);
+            if (Utils::stringEndsIn($class, $beforeClass)) {
```
snmaynard: Why have we added `stringEndsIn`? Saying you need to use the fully qualified class name seems sensible - but maybe there is something I'm missing?

src/Client.php
```diff
@@ -58,7 +58,7 @@ class Client
      *
      * @var \Bugsnag\Pipeline
      */
-    protected $pipeline;
+    public $pipeline;
```
snmaynard: Can we make this readonly so you cant set the pipeline variable to a string or a totally different array?

tests/PipelineTest.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+class PipelineTest extends TestCase
+{
+    public function testCallableAddedToPipeline()
+    {
+        $pipeline = new Pipeline();
+        $pipeline->pipe(new TestCallbackA());
+        $pipeline->execute([], function ($item) {
```
snmaynard: These tests seem overly complex to me. Why not make the tests append to a string? so the string at the end would spell out 'abc' or 'bac' if b was run first?

src/Pipeline.php
Cawllec: I initially preferred this option, but the three implementations that I was considering were:
- Straight matching - I don't like this personally as I think it's inconvenient and too prone to false negatives, although it would be the fastest and most accurate method.
- Split the string and classname by '//', then match the last sections against each other - I like this as it allows you to go as qualified as you'd like, but it's more complex.
- String match the classname against the end of the string - It again lets you qualify as much or as little as you'd like, but is faster than splitting and matching elements of an array.  I still like this as it's flexible but fast.

src/Client.php
Cawllec: There's no easy way to make something readonly in PHP it seems, see [here](https://stackoverflow.com/questions/2343790/how-to-implement-a-read-only-member-variable-in-php).

We could use the `__get` method override as described above, but as far as I can tell we'd then have to create entries for each class member, to ensure everything else gets passed back as required.

We could also use a getter method, which isn't nice, but would minimise the required changes and complexity.

kattrali: A getter method would be preferable, as pipeline should not be arbitrarily writeable. The same rationale applies to other Client properties like `$config`, which has `getConfig()`.

tests/PipelineTest.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function testInsertBeforeNoClass()
+    {
+        $pipeline = new Pipeline();
+        $pipeline->pipe(new TestCallbackA());
+        $pipeline->pipe(new TestCallbackC());
+        $pipeline->insertBefore(new TestCallbackB(), 'NotPresent');
+        $pipeline->execute('', function ($item) {
+            $this->assertSame(3, strlen($item));
```
snmaynard: The length checks are superfluous?

src/Utils.php
```diff
@@ -18,6 +18,27 @@ public static function functionAvailable($func)
...
 
+    /**
+     * Check a string ends in a substring.
+     *
+     * @param string $string The parent string
+     * @param string $sub The substring
+     *
+     * @return bool
+     */
+    public static function stringEndsIn($string, $sub)
```
GrahamCampbell: This function seems to be unnecessarily complex. I think this would be simpler:

```php
public static function stringEndsWith($haystack, $needle)
{
    return substr($haystack, -strlen($needle)) === $needle;
}
```

Also, the naming of parameters as haystack and needle would be consistent with PHP's standard API.

GrahamCampbell: If you want to support empty string as needle, you'll need to check if it's length is 0:

```php
public static function stringEndsWith($haystack, $needle)
{
    $length = strlen($needle);
    if ($length === 0) {
        return true;
    }

    return substr($haystack, -$length) === $needle;
}
```

Cawllec: That does make more sense, thanks

src/Pipeline.php
snmaynard: We should make this straight matching imo

PR #505, opened 2018-12-07
ci: Remove hhvm test target

kattrali:
HHVM has gone its own way from the true path of PHP 7

Ref: https://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,29 +28,8 @@ matrix:
       env: GUZZLE_VERSION=^5.3
     - php: 7.2
       env: GUZZLE_VERSION=^6.0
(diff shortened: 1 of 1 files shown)


Cawllec:
We can remove the constraint on mockery from [here](https://github.com/bugsnag/bugsnag-php/pull/499/files) at the same time.

GrahamCampbell:
Is there any stats on how many reports are being sent to Bugsnag from HHVM?

kattrali:
Not much in the way of stats, though its something we are working on. Its literally never come up in a request or support though.

PR #507, opened 2018-12-07
fix: Allow modification of stack values by using references

kattrali:
Makes it possible to modify stack frame values without having to
completely reconstruct the stack from modified frames.

For example, this now works:

```php
\Bugsnag::registerCallback(function ($report) {
    $frames = &$report->getStacktrace()->getFrames();
    foreach ($frames as &$frame) {
        $frame["inProject"] = true;
    }
});
```

## Changeset

Added reference indicators to `toArray()` and `getFrames()`

## Tests

* Added a new callback test which changes a frame value by reference

## Review

  - [x] Final review

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -167,18 +167,18 @@ class ClientTest extends TestCase
 
         $this->client->notify($report = Report::fromNamedError($this->config, 'Magic', 'oh no'));
 
-        $this->assertFalse($report->getStacktrace()->getFrames()[0]["inProject"]);
+        $this->assertFalse($report->getStacktrace()->getFrames()[0]['inProject']);
 
         $this->client->registerCallback(function (Report $report) {
             $frames = &$report->getStacktrace()->getFrames();
-            $frames[0]["inProject"] = true;
+            $frames[0]['inProject'] = true;
 
             return true;
         });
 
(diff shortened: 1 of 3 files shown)


tests/ClientTest.php
```diff
@@ -159,6 +159,28 @@ public function testBeforeNotifySkipsError()
...
+
+        $this->client->setBatchSending(false);
+
+        $this->client->notify($report = Report::fromNamedError($this->config, 'Magic', 'oh no'));
+
+        $this->assertFalse($report->getStacktrace()->getFrames()[0]['inProject']);
+
+        $this->client->registerCallback(function (Report $report) {
+            $frames = &$report->getStacktrace()->getFrames();
+            $frames[0]['inProject'] = true;
```
Cawllec: Can we test against something that explicitly won't be there, as `inProject` could conceivably be `true`, making this a false positive (or at least ambiguous to read).  Something like:
`$frames[0]['testFlag'] = true;`
With a corresponding assertion?

kattrali: That's why there are two assertions in this test. One to check that the value is false by default and then adding a callback, notifying, and checking that the value is true.

GrahamCampbell:
This might lead to unintended side effects if people are modifying the result from the getter.

kattrali:
Don’t really see how in a practical sense; it would still need to be referenced correctly to modify the report.

PR #514, opened 2019-01-23
Add setProjectRootRegex method

Cawllec:
## Goal
Adds `setProjectRootRegex` to the `Configuration` class, allowing the `projectRoot` regex to be set directly.

Based on #501 by @jpcid, this adds additional tests and a dedicated method in the Client class to allow calling the added `setProjectRootRegex` method in the Configuration.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -282,6 +282,7 @@ class Configuration
         }
 
         $this->projectRootRegex = $projectRootRegex;
+        $this->setStripPathRegex($projectRootRegex);
     }
 
     /**
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -261,7 +261,8 @@ class Configuration
      */
     public function setProjectRoot($projectRoot)
     {
-        $this->setProjectRootRegex('/^'.preg_quote($projectRoot, '/').'[\\/]?/i');
+        $projectRootRegex = $projectRoot ? '/^'.preg_quote($projectRoot, '/').'[\\/]?/i' : null;
+        $this->setProjectRootRegex($projectRootRegex);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Configuration.php
```diff
@@ -268,6 +268,22 @@ public function setProjectRoot($projectRoot)
...
+     * @return void
+     */
+    public function setProjectRootRegex($projectRootRegex)
+    {
+        if (@preg_match($projectRootRegex, null) === false) {
+            throw new InvalidArgumentException('Invalid project root regex: '.$projectRootRegex);
+        }
+
+        $this->projectRootRegex = $projectRootRegex;
+    }
```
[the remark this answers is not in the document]
Cawllec: I've added some which are basically copies of the existing client ones for `setStripPathRegex`, as well as testing the call through from `setProjectRootRegex`.

src/Configuration.php
```diff
@@ -261,11 +261,24 @@ public function getFilters()
      */
     public function setProjectRoot($projectRoot)
     {
-        $this->projectRootRegex = $projectRoot ? '/^'.preg_quote($projectRoot, '/').'[\\/]?/i' : null;
+        $this->setProjectRootRegex('/^'.preg_quote($projectRoot, '/').'[\\/]?/i');
```
kattrali: This should still check for and allow a null value to be set without throwing an exception in setProjectRootRegex.

Cawllec: Added some checks to enable the `stripPath` & `projectRoot` to be null.  As we already check for truthiness in the `isInProject` and `getStrippedFilePath` methods this will cover these cases.

jpcid:
You guys are awesome! I was very busy last month and I did not have time to apply your feedback to the original PR

Thank you @kattrali and @Cawllec.

PR #517, opened 2019-02-06
Auto on session tracking

Cawllec:
## Goal

- Allow endpoints to be set through the `Client`/`Configuration` after initialization
- Ensure only one GuzzleClient is created during each run
- Ensure GuzzleClient is only created if necessary
- Make AutoCaptureSessions `true` by default

TODO: Add SessionTracker tests

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -16,13 +16,6 @@ class HttpClient
     protected $config;
 
     /**
-     * The guzzle client instance.
-     *
-     * @var \GuzzleHttp\ClientInterface
-     */
-    protected $guzzle;
-
-    /**
      * The queue of reports to send.
      *
      * @var \Bugsnag\Report[]
@@ -47,14 +40,12 @@ class HttpClient
      * Create a new http client instance.
      *
      * @param \Bugsnag\Configuration      $config the configuration instance
-     * @param \GuzzleHttp\ClientInterface $guzzle the guzzle client instance
      *
      * @return void
      */
-    public function __construct(Configuration $config, ClientInterface $guzzle)
+    public function __construct(Configuration $config)
     {
         $this->config = $config;
-        $this->guzzle = $guzzle;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


UPGRADING.md
```diff
@@ -1,14 +1,27 @@
 Upgrading
 =========
 
+## 3.x to 4.x
+
+#### Setting Endpoints
+
+The way endpoints will be configured has been change slightly, in order to make sure all of your requests get send to the correct place.  You'll no longer be able to pass the `endpoint` parameter to your `Bugsnag\Client::make` calls, and should instead use the [`setEndpoints` configuration option](https://docs.bugsnag.com/platforms/php/other/configuration-options/#endpoints).  If you've previously used the `BUGSNAG_ENDPOINT` environment variable, it's now been renamed to `BUGSNAG_NOTIFY_ENDPOINT`, and will only work in conjunction with the `BUGSNAG_SESSION_ENDPOINT` environment variable, to make sure that both endpoints are being used correctly.
```
kattrali: A code sample to show a diff between how it used to work and how it works now would be helpful.

```diff
- old way
+ new way
```

UPGRADING.md
```diff
@@ -1,14 +1,78 @@
...
+
+```
+$bugnsag = Bugsnag\Client::make("YOUR_API_KEY");
+$bugsnag->setEndpoints(
+    "http://bugsnag-notify.example.com",
+    "http://bugsnag-session.example.com"
+);
+```
+
+#### Configuring the GuzzleHttp Client
```
snmaynard: Why is this change required for always on session tracking?

UPGRADING.md
```diff
@@ -1,14 +1,78 @@
...
+```
+
+New way:
+
+```
+$bugsnag = new Bugsnag\Client::make("YOUR_API_KEY");
+$bugsnag->setGuzzleClient($myCustomGuzzle);
+```
+
+#### Deprecation
```
snmaynard: Why is this change required for always on session tracking?

UPGRADING.md
```diff
@@ -1,14 +1,78 @@
 Upgrading
 =========
 
+## 3.x to 4.x
```
snmaynard: Rather than bumping the major version why not support backwards compat?

src/HttpClient.php
```diff
@@ -47,14 +40,12 @@ class HttpClient
      * Create a new http client instance.
      *
      * @param \Bugsnag\Configuration      $config the configuration instance
-     * @param \GuzzleHttp\ClientInterface $guzzle the guzzle client instance
      *
      * @return void
      */
-    public function __construct(Configuration $config, ClientInterface $guzzle)
+    public function __construct(Configuration $config)
```
snmaynard: why has passing guzzle in been deprecated?

Cawllec:
I've removed the non-pertinent changes that were a part of this PR originally, and attempted to clarify a number of the questions that require more context.

UPGRADING.md
Cawllec: This isn't, but should be removed _if_ this is a major release.  I've removed it as it'll be a part of a release PR instead.

Cawllec:
This is on hold for now, and will require review and possibly redevelopment in the future

PR #520, opened 2019-02-13
SessionTracker tests

Cawllec:
## Goal
Adds unit tests to the SessionTracker.

PR pointed toward AOST branch for now, as it derives from there.

PR #521, opened 2019-02-18
Fix/session guzzle client

Cawllec:
## Goal

Enables using the same GuzzleHttp\Client for all HTTP requests, meaning any custom options set will be applicable across event, session, & build notifications.

## Design

- Adds a public method `deliverPayload` to the HttpClient
- Changes SessionTracker to use `deliverPayload` for session delivery
- Adds `notifyEndpoint` to configuration with appropriate getter & setters
- Deprecate `getSessionClient` method on Configuration and Client
- Extracts `base_uri`/`base_url` from passed guzzle clients and assigns the value to `notifyEndpoint` to retain existing behaviour.

## Tests

- Updates existing tests to work with any API changes:
    - Removes reflection into `HttpClient` when testing default client setup
    - Guzzle client now expects blank headers when delivery build or deploy notifications
- Adds tests and lines in existing tests referencing new behaviour.
- Includes updates for tests established in #520 , requiring that to be merged first.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -115,7 +115,7 @@ class Client
         $this->recorder = new Recorder();
         $this->pipeline = new Pipeline();
         $this->http = new HttpClient($config, $guzzle ?: static::makeGuzzle());
-        $this->sessionTracker = new SessionTracker($config);
+        $this->sessionTracker = new SessionTracker($config, $this->http);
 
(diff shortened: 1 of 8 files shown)


imjoehaines:
Continued in https://github.com/bugsnag/bugsnag-php/pull/587

PR #522, opened 2019-04-01
Conditionally register shutdown function

ryancastle:
Don't call `register_shutdown_function()` unless `batchSending` is enabled.

## Goal
Prevent memory leaks when `Client` is constructed more than once in a process.

## Changeset

Added conditional clause to constructor, to prevent unnecessarily registering shutdown functions.

## Tests

Tested manually.

## Discussion

Each time `register_shutdown_function([$this, 'flush'])` is run, PHP encloses the state of the Bugsnag client until the process terminates. If this happens in a long-running process like phpunit, memory consumption increases continually. 

Disabling this feature unless batch sending is enabled will at least give users the option to prevent the memory leak. Also, in a testing environment enabling Bugsnag with `batchSending=false` may make sense in some contexts, but it's unlikely that `batchSending=true` would ever be appropriate.

### Alternative Approaches

Stubbing the client and factory and updating configuration based on context.

### Outstanding Questions

Additional changes to `ClientFactory` are required in `bugsnag/bugsnag-symfony` to ensure that client configuration is properly bootstrapped.

```
$config = new Config($this->key ?: '');
$config->setBatchSending($this->batch);

$client = new Client($config, $this->resolver, $guzzle);
```
### Linked issues

Fixes https://github.com/bugsnag/bugsnag-symfony/issues/77

## Review

- [ ] Commented on code changes inline explain the reasoning behind the approach
  - [ ] Final review

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -121,7 +121,9 @@ class Client
         $this->registerMiddleware(new BreadcrumbData($this->recorder));
         $this->registerMiddleware(new SessionData($this));
 
-        register_shutdown_function([$this, 'flush']);
+        if ($this->config->isBatchSending()) {
+            register_shutdown_function([$this, 'flush']);
+        }
     }
 
     /**


phillipsam:
Fixed in [v1.6.0](https://github.com/bugsnag/bugsnag-symfony/releases/tag/v1.6.0)

PR #525, opened 2019-05-21
Add PHP runtime version reporting

tomlongridge:
## Goal

Augments reporting with the version of PHP in use when an event occurred in case of issues in a specific version of PHP.

## Design

Added data to device.runtimeVersions.php as per other notifiers.

### Changed

* Configuration.php - added the data during initialisation;
* Configuration.php - added mergeDeviceData method for use by other PHP notifiers.

## Tests

* Added testMergeDeviceData* unit tests
* Amended assertion structure of testDeviceData
* End-to-end test run against live project ([event here](https://app.bugsnag.com/bugsnag-test-projects/php/errors/5ce408ae0b9f150019679df1?event_id=5ce40cbc003a00195b430000))

## Review

  - [x] Final review

- [x] Consistency across platforms for structures or concepts added or modified
- [x] Consistency between the changeset and the goal stated above
- [x] Internal consistency with the rest of the library - is there any overlap between existing interfaces and any which have been added?
- [x] Usage friction - is the proposed change in usage cumbersome or complicated?
- [x] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?
- [x] Concurrency concerns - if components are accessed asynchronously, what issues will arise
- [x] Thoroughness of added tests and any missing edge cases
- [x] Idiomatic use of the language

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -704,18 +704,6 @@ class Client
     }
 
     /**
-    * Adds new data fields to the device data collection.
-    *
-    * @param array $data an associative array containing the new data to be added
-    *
-    * @return this
-    */
-    public function mergeDeviceData($data)
-    {
-        return $this->config->mergeDeviceData($data);
-    }
-
-    /**
      * Get the device data.
      *
      * @return array
--- a/src/Client.php
+++ b/src/Client.php
@@ -704,6 +704,18 @@ class Client
     }
 
     /**
+    * Adds new data fields to the device data collection.
+    *
+    * @param array $data an associative array containing the new data to be added
+    *
+    * @return this
+    */
+    public function mergeDeviceData($data)
+    {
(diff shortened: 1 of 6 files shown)


src/Client.php
```diff
@@ -703,6 +703,18 @@ public function setHostname($hostname)
...
     }
 
+    /**
+    * Adds new data fields to the device data collection.
+    *
+    * @param array $data an associative array containing the new data to be added
+    *
+    * @return this
+    */
+    public function mergeDeviceData($data)
```
kattrali: Forgot one thing: This isn't needed here. For public usage, `$client->getConfig()->mergeDeviceData()` is available. Older methods were copied to avoid making a breaking change during #510, but since this is new, there's no need to have duplicate methods.

PR #538, opened 2019-08-21
Bump httpclient allowed payload limit

qwertyuu:
## Goal

Bring the upload limit closer to the actual Bugsnag's error reporting limit, see: https://github.com/bugsnag/bugsnag-dotnet/blob/96360fb0f8d31f5fd6a8f3f17184bd626421820e/src/Bugsnag/Payload/Report.cs#L16

from `bugsnag-dotnet`. Also here in the documentation: https://docs.bugsnag.com/platforms/dotnet/asp-net/configuration-options/#maximum-breadcrumbs

### Changed

Changed `MAX_SIZE` const in `HttpClient` to 1024*1024, or 1048576.

## Tests

Used the new constant to ensure that "massive payloads" are properly tested if the value ever changes.

## Review

  - [x] Final review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-## 3.17.1 (2019-08-21)
+## TBD
 
 Bump max payload limit to follow Bugsnag's backend limit to enable larger errors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+## 3.17.1 (2019-08-21)
+
+Bump max payload limit to follow Bugsnag's backend limit to enable larger errors
+
+### Enhancements
+
+* Bump max payload limit to follow Bugsnag's backend limit to enable larger errors
+  [#538](https://github.com/bugsnag/bugsnag-php/pull/538)
+
 ## 3.17.0 (2019-05-28)
 
 Add the version of PHP to report information to allow easier diagnosis of version-specific issues.
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -34,7 +34,7 @@ class HttpClient
      *
      * @var int
      */
-    const MAX_SIZE = 1024 * 1024;
+    const MAX_SIZE = 1048576;
 
     /**
      * The current payload version.
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -34,7 +34,7 @@ class HttpClient
      *
      * @var int
      */
-    const MAX_SIZE = 500000;
+    const MAX_SIZE = 1048576;
 
     /**
      * The current payload version.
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
@kattrali I wonder if the payload limit was 512kb in 2015, which would explain why the code is like this?

src/HttpClient.php
```diff
@@ -34,7 +34,7 @@ class HttpClient
      *
      * @var int
      */
-    const MAX_SIZE = 500000;
+    const MAX_SIZE = 1048576;
```
kattrali: This could instead be the calculation for the value, so its obvious at first glance what is happening here.

```suggestion
    const MAX_SIZE = 1024 * 1024;
```

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+## 3.17.1 (2019-08-21)
```
kattrali: No need to add a date, we put it in when we do the release.
```suggestion
## TBD
```

tests/HttpClientTest.php
```diff
@@ -79,7 +79,7 @@ public function testMassiveMetaDataHttpClient()
         $this->guzzle->expects($spy = $this->any())->method('post');
 
         // Add a report to the http and deliver it
-        $this->http->queue(Report::fromNamedError($this->config, 'Name')->setMetaData(['foo' => str_repeat('A', 1000000)]));
+        $this->http->queue(Report::fromNamedError($this->config, 'Name')->setMetaData(['foo' => str_repeat('A', HttpClient::MAX_SIZE * 2)]));
```
kattrali: This test (and the others) should still use a constant value to verify that the proper checks are happening and that `MAX_SIZE` has not been unexpectedly altered.
```suggestion
        $this->http->queue(Report::fromNamedError($this->config, 'Name')->setMetaData(['foo' => str_repeat('A', 1500000)]));
```

qwertyuu:
@kattrali 👍

qwertyuu:
@kattrali 👍

src/HttpClient.php
qwertyuu: PHP 5.5 does not seem to like that I use an expression instead of a constant value. https://travis-ci.com/bugsnag/bugsnag-php/jobs/227394423

I'll need to keep it as a whole number for as long as we support that PHP version in the CI.

GrahamCampbell: Yeh, only 5.6+ allow a wider class of expessions as constants, unfortunately.

kattrali: I’d forgotten that it might anger The Ancient Ones. Well maybe add a comment to say it’s a Mb.

qwertyuu:
@kattrali Comment added!

kattrali:
Thanks for the quick edits! ⚡️

PR #540, opened 2019-08-31
Compatibility with phpdotenv v2

GrahamCampbell:
We should probably not insist on v3, to allow the latest version of the bugsnag php library to be installed in every application.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php": ">=5.5",
         "composer/ca-bundle": "^1.0",
         "guzzlehttp/guzzle": "^5.0|^6.0",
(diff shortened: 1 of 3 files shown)


abigail-bugsnag:
Hi @GrahamCampbell - thanks for this! We will look to get this merged.

GrahamCampbell:
Any movement here? This change is needed so the latest version of Bugsnag can be installed in Laravel 5, which depends on phpdotenv v2.

rjharrison:
Thanks Graham. We've fixed via https://github.com/bugsnag/bugsnag-php/pull/551.

Issue #523, opened 2019-04-04
Bugsnag exception handler causes CLI scripts to exit 0 with uncaught exceptions


blkperl:
### Description

Bugsnag exception handler causes CLI scripts to exit 0 with uncaught exceptions

### Issue

Bugsnag exception handler causes CLI scripts to exit 0 with uncaught exceptions

### Environment

* bugsnag-php version: v3.16.0
* PHP version: 7.2
* What web framework are you using, if any?: None
* How are you installing bugsnag-php? (Composer, phar files?): composer
* Operating system name and version: Ubuntu 16.04

#### How to reproduce the issue

```php
<?php
declare(strict_types=1);

require __DIR__ . '/vendor/autoload.php';


$bugsnagClient = \Bugsnag\Client::make('foo');
\Bugsnag\Handler::registerWithPrevious($bugsnagClient);
throw new Exception();
```

```
$ php test.php
$ echo $?
0
```

#### How to workaround the issue

```php
$ cat test.php
<?php
declare(strict_types=1);

require __DIR__ . '/vendor/autoload.php';


$bugsnagClient = \Bugsnag\Client::make('foo');

set_exception_handler(function ($throwable) {
      throw $throwable;
});

\Bugsnag\Handler::registerWithPrevious($bugsnagClient);
throw new Exception();
```

```
$ php test.php
PHP Fatal error:  Uncaught Exception in test.php:14
Stack trace:
#0 {main}
  thrown in test.php on line 14

Fatal error: Uncaught Exception in test.php on line 14

Exception:  in test.php on line 14

Call Stack:
    0.0240    2079608   1. Bugsnag\Handler->exceptionHandler() REDACTED/vendor/bugsnag/bugsnag/src/Handler.php:0
    0.0282    2200056   2. call_user_func:{REDACTED/vendor/bugsnag/bugsnag/src/Handler.php:155}() REDACTED/vendor/bugsnag/bugsnag/src/Handler.php:155
    0.0282    2200056   3. {closure:test.php:9-11}() REDACTED/vendor/bugsnag/bugsnag/src/Handler.php:155

$ echo $?
255
```

glennpratt:
I have the same issue, this has now bitten me multiple times.

xander-jones:
Hi @glennpratt, have you tried using the workaround in @blkperl's issue report using `set_exception_handler()`? You may also be interested in reading the comments in #475 as to why this is happening. --Xander

glennpratt:
[...]

Issue #475, opened 2018-05-01
Reporting unhandled exceptions breaks error page handling


thedotedge:
### Expected behavior
When uncaught exception is thrown, bugsnag intercepts it without possibility to capture it and show custom error page to the user.

### Steps to reproduce
```php
<?php
// some code that emits E_NOTICE
// ...

register_shutdown_function(function () use ($bugsnag) {
        if ($error = error_get_last()) {
            if (in_array($error['type'], [E_ERROR, E_PARSE, E_COMPILE_ERROR, E_CORE_ERROR, E_USER_ERROR])) {
                http_response_code(500);
                exit();
            }
        }
    });

$bugsnag = Bugsnag\Client::make($GLOBALS['config']['bugsnag']['api_key']);
Bugsnag\Handler::register($bugsnag);

throw new \RuntimeException('Triggered error 500');
```

1. When bugsnag handler is not registered, response has status code 500 error_get_last() correctly returns  E_ERROR (intended behaviour)
2. When bugsnag handler is registered, error_get_last() does not return E_ERROR from RuntimeException anymore, rather some E_NOTICE from prior code.

### Version
bugsnag-php v3.12.1
php-fpm 7.2.2

snmaynard:
I believe you can call `registerWithPrevious` as in https://github.com/bugsnag/bugsnag-php/blob/master/src/Handler.php#L51

We should improve the documentation here at a minimum. But it feels like that should be the default behaviour. cc @kattrali @Cawllec

Cawllec:
Yes, you can call `registerWithPrevious` in order to ensure the previous handlers aren't overwritten.  I'll update the docs as I can't see where we mention this.

thedotedge:
Furthermore, bugsnag somehow messes up response code in case of fatal error:

This returns status code 500:
```
<?php
ini_set('display_errors', 0);
throw new \RuntimeException('Triggered error 500'); 
```
This returns status code 200:
```
ini_set('display_errors', 0);
$bugsnag = Bugsnag\Client::make($GLOBALS['config']['bugsnag']['api_key']);
Bugsnag\Handler::register($bugsnag);
throw new \RuntimeException('Triggered error 500'); 
```

Cawllec:
I'm not in the office the rest of this week, but will get to this as soon as possible once I'm back.

thedotedge:
Any updates @Cawllec?

Cawllec:
[...]

PR #541, opened 2019-09-02
Default PHP exception handler will now fire

rjharrison:
This is a PR for [issue 523](https://github.com/bugsnag/bugsnag-php/issues/523) (And accidentally fixes [issue 475](https://github.com/bugsnag/bugsnag-php/issues/475))

In both cases, the root cause is that BugSnag **does** change default PHP behaviour with regard to exception handling and `registerWithPrevious` when no previous exception handler has been set.

It is more obvious in the CLI context, where the exit status from a script encountering an unhandled exception is incorrectly 0 ([issue 523](https://github.com/bugsnag/bugsnag-php/issues/523)). Without BugSnag enabled, the default PHP behaviour would be to generate a fatal error (`E_ERROR`) and the CLI script would exit with 255.

In a SAPI context, this is less obvious. Assuming `display_errors = off` and no content has been sent to the browser (ie output buffering is enabled), PHP default behaviour is to show a white screen of death with an HTTP status code of **500**. With BugSnag enabled via `registerWithPrevious`, the behaviour is to display a white screen of death with an HTTP status code of **200**. This is effectively the same thing as the CLI bug, but it's rarely noticeable unless something (a proxy or upstream server like Apache) interrogates the HTTP status code.

The additional effect that @thedotedge reported in [issue 475](https://github.com/bugsnag/bugsnag-php/issues/475) is that the default behaviour produces a fatal error that can be "trapped" (ish) within the shutdown function. When BugSnag is enabled via `registerWithPrevious` this error is never created and his switch statement fails (he makes a legitimate point that this is a bug).

In summary, if no previous exception handler exists, and the user invokes BugSnag via `registerWithPrevious`, then BugSnag should trigger the default exception handler in PHP. 

Note: a side effect to this is that BugSnag will now report on two events, the unhandled exception and the resulting fatal error. Perhaps the `E_ERROR` is something that could be filtered on the reporting side of things (the timestamp looks like it's the same for both events?)
Screenshot 2019-09-02 at 22 04 01

### DIFF
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -26,6 +26,12 @@ class Handler
     protected $previousExceptionHandler;
 
     /**
+     * Indicates that we should use the previous exception handler
+     * @var boolean
+     */
+    protected $usePreviousExceptionHandler;
+
+    /**
      * Register our handlers.
      *
      * @param \Bugsnag\Client|string|null $client client instance or api key
@@ -100,6 +106,7 @@ class Handler
 
         if ($callPrevious) {
             $this->previousExceptionHandler = $previous;
+            $this->usePreviousExceptionHandler = true;
         }
     }
 
@@ -149,11 +156,19 @@ class Handler
 
         $this->client->notify($report);
 
-        if ($this->previousExceptionHandler) {
-            call_user_func(
-                $this->previousExceptionHandler,
-                $throwable
-            );
+        if ($this->usePreviousExceptionHandler) {
+            // if there is a callable...
+            if ($this->previousExceptionHandler) {
+                call_user_func(
+                    $this->previousExceptionHandler,
+                    $throwable
+                );
+            } else {
+                // There was no previous exception handler registered.
+                // In this case, restore to the default handler and re-throw
+                restore_exception_handler();
+                throw $throwable;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR. My only concern is that this may break a lot of existing code.

rjharrison:
You know the product usage better than me, but at the point of an unhandled exception, without any exception handlers already registered, the PHP app is blowing up beyond recovery - it's already pretty broken. Will "the thing that called the PHP" break with this change? Maybe yea - but I can't imagine there are many people relying on this behaviour intentionally. I think it's more likely that if you need to react to a fatal error outside of PHP (ie Apache http redirection/serving static error pages/reacting via CLI) that you'd try and work around this bug (like forcing a non-zero exit status on the CLI). 

Good luck with it :-)

abigail-bugsnag:
@rjharrison thanks for your comment. We agree with @GrahamCampbell that this will probably have a significant impact, however we also agree that the current behaviour doesn't seem correct.
We will discuss this and get back to you.

imjoehaines:
Thanks for the PR, @rjharrison 

I've just released [v3.23.1](https://github.com/bugsnag/bugsnag-php/releases/tag/v3.23.1) which should resolve this issue, as well as some other issues with the error/exception handler

PR #544, opened 2019-10-24
Remove Breadcrumb "name" length restriction

rjharrison:
## Goal

Breadcrumb "name" field is currently capped at 30-characters, anything larger is truncated. This PR removes this restriction and allows any length string to be sent from the client. 

https://bugsnag.atlassian.net/browse/PLAT-3671

### Removed

- strlen() calls in the Breadcrumb constructor + Client::leaveBreadcrumb() function
- Breadcrumb::MAX_LENGTH constant no longer used

## Tests
Removed tests that checked for 30-character max lenth

### Outstanding Questions

It felt a little odd removing a max-length restriction and leaving nothing in its place. I wondered what happens if a client sends a massive string... but I guess they could do that manually via the API, so the client library doesn't protect against that sort of thing.

Also wondered if there _really_ is no max-length any more? 

## Review

- [ ] Commented on code changes inline explain the reasoning behind the approach
  - [x] Final review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## (TBD)
+
+### Enhancements
+
+* Removed the length restriction on the Breadcrumb "name" field to match the API and allow large names to be sent.
+
 ## 3.18.0 (2019-08-28)
 
 ### Enhancements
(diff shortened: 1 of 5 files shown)


kattrali:
@GrahamCampbell The layout of breadcrumbs in the UI has changed, allowing the name to function as a long message. This change is to accommodate that.

PR #547, opened 2019-10-28
Add shutdown strategy

rjharrison:
## Goal

We currently rely on register_shutdown_function() to defer calls to flush(). This potentially causes a [memory leak](https://github.com/bugsnag/bugsnag-symfony/issues/77) as references to the Client are held until the script exits. (This is only an issue when testing with many instances of Client objects). 

The call to register_shutdown_function is hardcoded in the constructor. This PR introduces a ShutdownStrategyInterface which replaces this hardcoding and allows for the injection of different shutdown strategies. 

There is no functionality change in this PR. The current register_shutdown_function() behaviour is now wrapped inside an object. In a future PR on the bugsnag-symfony project I will introduce a Symfony-specific shutdown strategy that will fix [issue-77](https://github.com/bugsnag/bugsnag-symfony/issues/77).

### Added

- New ShutdownStrategyInterface parameter to Client constructor (defaults to null)
- PhpShutdownStrategy (replicates current behaviour)

### Removed

- hardcoded call to register_shutdown_function

## Tests

Added unit tests. The "shutdownSpy" is a pretty cool way to test native functions (in this case the call to register_shutdown_function)

### Alternative Approaches

None.

### Outstanding Questions

I think we should also make the call to registerShutdownStrategy conditional on whether we are in batch mode or not (there is no need to defer calls to flush() if we are not "batching"). However at this stage I want to change one thing at a time, so will potentially do this as a separate piece of work.

## Review

  - [x] Final review

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -128,7 +128,7 @@ class Client
 
         // Shutdown strategy is used to trigger flush() calls when batch sending is enabled
         $shutdownStrategy = $shutdownStrategy ?: new PhpShutdownStrategy();
-        $shutdownStrategy->register($this);
+        $shutdownStrategy->registerShutdownStrategy($this);
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -106,10 +106,12 @@ class Client
     /**
      * Create a new client instance.
      *
-     * @param \Bugsnag\Configuration $config
+     * @param \Bugsnag\Configuration                  $config
      * @param \Bugsnag\Request\ResolverInterface|null $resolver
-     * @param \GuzzleHttp\ClientInterface|null $guzzle
-     * @param ShutdownStrategyInterface $shutdownStrategy
+     * @param \GuzzleHttp\ClientInterface|null        $guzzle
+     * @param ShutdownStrategyInterface               $shutdownStrategy
+     * 
+     * @return void
      */
     public function __construct(Configuration $config, ResolverInterface $resolver = null, ClientInterface $guzzle = null, ShutdownStrategyInterface $shutdownStrategy = null)
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -106,10 +106,10 @@ class Client
     /**
      * Create a new client instance.
      *
-     * @param \Bugsnag\Configuration                  $config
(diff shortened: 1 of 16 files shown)


src/Client.php
```diff
@@ -107,10 +109,11 @@ public static function make($apiKey = null, $endpoint = null, $defaults = true)
      * @param \Bugsnag\Configuration                  $config
      * @param \Bugsnag\Request\ResolverInterface|null $resolver
      * @param \GuzzleHttp\ClientInterface|null        $guzzle
+     * @param ShutdownStrategyInterface               $shutdownStrategy
```
GrahamCampbell: Missing `|null`.

GrahamCampbell: Also, should use a FQCN out of convention.

src/Shutdown/ShutdownStrategyInterface.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Interface ShutdownStrategyInterface.
+ */
+interface ShutdownStrategyInterface
+{
+    /**
+     * Register the shutdown behaviour.
+     *
+     * @param Client $client
+     */
```
GrahamCampbell: Should use a FQCN. Also, missing a return void annotations to let people know that they are not expected to return anything, Vs undocumented.

src/Client.php
rjharrison: Fixed, thanks.

src/Shutdown/ShutdownStrategyInterface.php
rjharrison: Fixed, thanks.

src/Shutdown/ShutdownStrategyInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Bugsnag\Shutdown;
+
+use Bugsnag\Client;
+
+/**
+ * Interface ShutdownStrategyInterface.
```
tomlongridge: Worth expanding the docs here to describe the purpose of the interface and expected reasons for implementing it?

PR #553, opened 2019-12-03
Test on PHP 7.4 too

GrahamCampbell:


GrahamCampbell:
Looks like there are some issues with PHP 7.4.

Cawllec:
It looks like this is due to using an older version of PHPUnit, and possibly because we've not had `convertWarningsToExceptions` enabled before.  Looks like updating it will take updating PHP-mock and the tests using `PHPUnit_Framework_TestCase` as well.

Cawllec:
@tomlongridge I'd recommend we merge #552 for now (adding 7.3 tests) and ticket 7.4 tests.

Cawllec:
On hold for now until the PHPUnit version issue is resolved.

GrahamCampbell:
> @tomlongridge I'd recommend we merge #552 for now (adding 7.3 tests)

I agree, the 7.3 PR can be merged without this, and should be.

GrahamCampbell:
Actually, this seems to be caused by Mockery, not PHPUnit... I think! Going to try fiddling with the requirements...

GrahamCampbell:
Oh. It was both phpunit and mockery. Well, I fixed the mockery related error anyway...

PR #556, opened 2019-12-20
Support PHP 7.4

GrahamCampbell:
This updates the tests so they can run on a newer version of PHPUnit as well as the old version, so we can continue to support PHP 5. This then allows us to run the test suite on PHP 7.4 without issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,54 @@
 language: php
-dist: trusty
-sudo: false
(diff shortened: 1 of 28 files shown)


GrahamCampbell:
Ready for review now. :)

phillipsam:
@GrahamCampbell thanks! :)

PR #567, opened 2020-02-14
Support Guzzle 7

GrahamCampbell:


GrahamCampbell:
Yes. I'll prep a PR with the change log stuff, which we can merge just before we tag the next release... whenever that will be. Probably late next week, or early March.

PR #573, opened 2020-02-26
Release v3.20.0

GrahamCampbell:
I'm treating support for newer PHP versions as a bug fix, like Symfony do.


GrahamCampbell:
It looks like the packing docs are not so good for 2020. We need to make sure we install PHP 5.5 compatible dependencies. Running composer on PHP 7 will be no good for building the phar with.

GrahamCampbell:
I'll see if I can update the script to by self contained and use PHP 5.5 from docker. There is no official PHP 5.5 docker image, but I'm sure there will be a well-ish used one somewhere we can use. Failing that, we can make our own Dockerfile.

GrahamCampbell:
So, actually, it's not such a big problem as i thought. I forgot that we don't pack our dependencies in the phar. Still, it makes sense to have a reproducible way to build the phar, which doesn't depend on local PHP config.

GrahamCampbell:
The build phar is available here:

[bugsnag.zip](https://github.com/bugsnag/bugsnag-php/files/4255472/bugsnag.zip)

GrahamCampbell:
The current contribution guide is still correct. You just run `make package`.

GrahamCampbell:
If our PHP 5.5 image has already been built, then we re-use it. Otherwise we build it from scratch.

Cawllec:
Is there a reason we couldn't use [this PHP image](https://hub.docker.com/layers/php/library/php/5.5.38-cli/images/sha256-15008d2bf3fbdf421b133e085dafb8f8640e8b20be4068d55ebf01dc211868b8?context=explore) instead?  There's still some minor adjustment required, but we can drop the dockerfile down to just:
```
FROM php:5.5.38-cli

ENV COMPOSER_ALLOW_SUPERUSER 1
RUN curl -sS https://getcomposer.org/installer | php \
    && mv composer.phar /usr/local/bin/composer
RUN apt-get update && apt-get install -y git zip
```
Which seems a fair bit faster as well.

GrahamCampbell:
Yes, I suppose so. The reason for me not doing this originally was because that image is full of security vulnerabilities (everything since the last 4 years), however, I wasn't able to get the image to build against a new version of alpine linux, so I think your suggestion is what we should go with.

PR #574, opened 2020-03-09
Add meta data interface and trait

migo315:
## Goal
Provide a very simple way to extend an exception with additional meta data everywhere without further dependencies.

## Design
This feature is very pragmatic and straightforward. Similar to [PSR Logger](https://github.com/php-fig/log/tree/master/Psr/Log), we provide a `MetaDataAwareInterface` and a `MetaDataAwareTrait`. The developer can (but must not) extend his Exceptions with the Interface / Trait to add some additional informations:

``` 
class UnrecoverableProcessException extends Exception implements MetaDataAwareInterface
{
    use MetaDataAwareTrait;
}
```

Later just pass your informations:
```
$exception = new UnrecoverableProcessException('I am an Exception');
$exception->addMetaData(['myTab' => [
    'foo' => 'bar',
]]);

throw $exception;
```

This implementation does not need any services or heavy structures.

### Added
- MetaDataAwareInterface
- MetaDataAwareTrait

### Changed
- Client add MetaData if Exception implements interface

### Linked issues
Original PR: https://github.com/bugsnag/bugsnag-symfony/pull/83

## Review

- [] Reviewed the test cases added for completeness and possible points for discussion
  - [x] Final review

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -10,6 +10,7 @@ use Bugsnag\Callbacks\RequestCookies;
 use Bugsnag\Callbacks\RequestMetaData;
 use Bugsnag\Callbacks\RequestSession;
 use Bugsnag\Callbacks\RequestUser;
+use Bugsnag\MetaData\MetaDataAwareInterface;
 use Bugsnag\Middleware\BreadcrumbData;
 use Bugsnag\Middleware\CallbackBridge;
 use Bugsnag\Middleware\NotificationSkipper;
@@ -274,6 +275,10 @@ class Client
     public function notifyException($throwable, callable $callback = null)
     {
         $report = Report::fromPHPThrowable($this->config, $throwable);
+
+        if ($throwable instanceof MetaDataAwareInterface) {
+            $report->addMetaData($throwable->getMetaData());
+        }
 
         $this->notify($report, $callback);
     }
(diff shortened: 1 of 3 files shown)


src/Client.php
```diff
@@ -275,6 +276,10 @@ public function notifyException($throwable, callable $callback = null)
     {
         $report = Report::fromPHPThrowable($this->config, $throwable);
 
+        if ($throwable instanceof MetaDataAwareInterface) {
```
GrahamCampbell: This should probably be implemented by a middleware, rather than being baked into the client like this.

migo315: I'm not that familiar with this SDK. But the middleware doesn't get any exceptions, does it?

GrahamCampbell: Yeh, that is correct. Maybe we should change the report object to keep the actual exception around, so people can grab it (by means of a method on the report object that returns the exception or null).

johnkiely1:
This has now been fixed via php release V3.21.0
https://github.com/bugsnag/bugsnag-php/releases/tag/v3.21.0.
running composer update bugsnag/bugsnag will pull in latest version

We have added support for the getting the original error on which you can attach additional metadata.

PR #576, opened 2020-03-10
Added support for getting the "original error"

GrahamCampbell:
Closes #574. Closes https://github.com/bugsnag/bugsnag-symfony/pull/83.

### DIFF
--- a/src/Report.php
+++ b/src/Report.php
@@ -212,9 +212,9 @@ class Report
      *
      * @return \Throwable|array|null
      */
-    public function originalError()
+    public function getOriginalError()
     {
-        return $this->original;
+        return $this->originalError;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Report.php
```diff
@@ -23,6 +23,13 @@ class Report
      */
     protected $config;
 
+    /**
+     * The original error.
+     *
+     * @var \Throwable|array|null
+     */
+    protected $original;
```
tomlongridge, line 31: Apologies, in our discussion I meant the field to be `originalError` and the standard getter `getOriginalError`.

```suggestion
    protected $originalError;
```

src/Report.php
```diff
@@ -200,6 +207,16 @@ protected function __construct(Configuration $config)
...
 
+    /**
+     * Get the original error.
+     *
+     * @return \Throwable|array|null
+     */
+    public function originalError()
+    {
+        return $this->original;
+    }
```
tomlongridge, line 218:
```suggestion
    public function getOriginalError()
    {
        return $this->originalError;
    }
```

PR #578, opened 2020-04-02
Use full class names in breadcrumbs

GrahamCampbell:
Replaces https://github.com/bugsnag/bugsnag-php/pull/470. We can do this now, since the UI can now deal with much larger names and the limit on breadcrumb size was removed from the library in https://github.com/bugsnag/bugsnag-php/pull/544.


GrahamCampbell:
👍

PR #561, opened 2020-01-08
More robust session count checks

GrahamCampbell:

PR #579, opened 2020-04-20
Fix session tracker count raising a warning on PHP 7.2

### LINKED ISSUE #557: PHP 7.2: count(): Parameter must be an array or an object that implements Countable
### Description
*count(): Parameter must be an array or an object that implements Countable* warning is raised.

### Issue
```php
        $sessions = $this->getSessionCounts();
        $this->setSessionCounts([]);
        if (count($sessions) == 0) {
            return;
        }
```
in bugsnag/bugsnag/src/SessionTracker.php:285
### Environment

* bugsnag-php version: 3.19.0
* PHP version: 7.2

imjoehaines:
## Goal

Fix a warning being raised when a session storage function returns `null` on PHP versions >=7.2 — see #557

### Added

- More tests for `SessionTracker`

### Changed

- `SessionTracker::deliverSessions` now checks if the return value from the session storage function is an array before calling `count` on it
- `SessionTracker::incrementSessions` now checks the return value from the session storage function is an array before calling `array_key_exists` on it
- `SessionTracker::incrementSessions` also checks the return value for "last sent" is an integer before doing maths with it, though we specifically cast `null` to `0` to keep the previous behaviour with `null` values
    I think this is a safe change because our code should be the only place this value is set (`setLastSent`), so this should only be a change in behaviour for incorrectly implemented `storageFunction` callbacks
- `SessionTracker` didn't import `Exception` or `InvalidArgumentException`
    This means none of the `setXFunction` methods would throw if given invalid parameters (they'd crash due to no `Bugsnag\InvalidArgumentException` class existing instead)
    It also means we wouldn't catch exceptions when `deliverSessions` was called, so the retry functionality wasn't working correctly
- Some of the namespaces for returned Guzzle `Client` objects have been updated
- StyleCI is failing on master so the suggested fixes have been applied on this branch (in 1364246ff0316ebe4114c93b1059d0bdd20e34de)

## Tests

- I've added a test for non-array return values for `getSessionCounts` to make sure the fix is working — these will fail if the warning was still being raised
- I improved the tests to avoid mocking the `SessionTracker` class itself, as it's the class being tested
- I've also added a test for when `startSession` is called, which is mostly the same as `sendSessions` but the session is pre-populated

### Outstanding Questions

- The return value of `Configuration::getSessionClient` is `GuzzleHttp\ClientInterface` which no longer has a `post` method available (as of Guzzle v6)
    This means we need to check whether it exists or not, to know which version we're on and therefore which method to call
    We could change the return type to the concrete `GuzzleHttp\Client` class instead, which has a `post` method in v7, v6 and v5 of Guzzle

### Linked issues

Fixes #557 
Supersedes https://github.com/bugsnag/bugsnag-laravel/pull/378 and https://github.com/bugsnag/bugsnag-php/pull/561

## Review

  - [x] Structural review of the classes, functions, and properties modified
  - [x] Final review

### DIFF
--- a/src/SessionTracker.php
+++ b/src/SessionTracker.php
@@ -153,12 +153,18 @@ class SessionTracker
     }
 
     /**
-     * @return mixed
+     * @return array
      */
     public function getCurrentSession()
     {
         if (is_callable($this->sessionFunction)) {
-            return call_user_func($this->sessionFunction);
+            $currentSession = call_user_func($this->sessionFunction);
+
+            if (is_array($currentSession)) {
+                return $currentSession;
+            }
+
+            return [];
         }
 
         return $this->currentSession;
@@ -261,10 +267,6 @@ class SessionTracker
         try {
             $sessionCounts = $this->getSessionCounts();
 
-            if (!is_array($sessionCounts)) {
-                $sessionCounts = [];
-            }
-
             if (array_key_exists($minute, $sessionCounts)) {
                 $sessionCounts[$minute] += $count;
             } else {
@@ -279,11 +281,7 @@ class SessionTracker
 
             $lastSent = $this->getLastSent();
 
-            if ($lastSent === null) {
-                $lastSent = 0;
-            }
-
-            if ($deliver && is_int($lastSent) && ((time() - $lastSent) > self::$DELIVERY_INTERVAL)) {
+            if ($deliver && ((time() - $lastSent) > self::$DELIVERY_INTERVAL)) {
                 $this->deliverSessions();
             }
         } finally {
@@ -294,13 +292,20 @@ class SessionTracker
     }
 
     /**
-     * @return mixed
+     * @return array
      */
     protected function getSessionCounts()
     {
         if (is_callable($this->storageFunction)) {
-            return call_user_func($this->storageFunction, self::$SESSION_COUNTS_KEY);
+            $sessionCounts = call_user_func($this->storageFunction, self::$SESSION_COUNTS_KEY);
+
+            if (is_array($sessionCounts)) {
+                return $sessionCounts;
+            }
+
+            return [];
         }
+
 
         return $this->sessionCounts;
     }
@@ -325,11 +330,14 @@ class SessionTracker
     protected function trimOldestSessions()
     {
         $sessions = $this->getSessionCounts();
+
         uksort($sessions, function ($key) {
             return strtotime($key);
         });
+
         $sessions = array_reverse($sessions);
         $sessionCounts = array_slice($sessions, 0, self::$MAX_SESSION_COUNT);
+
         $this->setSessionCounts($sessionCounts);
     }
 
@@ -362,7 +370,7 @@ class SessionTracker
 
         $this->setSessionCounts([]);
 
-        if (!is_array($sessions) || count($sessions) === 0) {
+        if (count($sessions) === 0) {
             return;
         }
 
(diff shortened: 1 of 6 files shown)


src/SessionTracker.php
```diff
@@ -281,27 +286,36 @@ protected function constructPayload($sessions)
     protected function deliverSessions()
     {
         $sessions = $this->getSessionCounts();
+
         $this->setSessionCounts([]);
-        if (count($sessions) == 0) {
+
+        if (!is_array($sessions) || count($sessions) === 0) {
```
imjoehaines, line 292: This `!is_array` check fixes the issue. It doesn't seem worth adding `Countable` support here, but we can do if that's desirable

src/SessionTracker.php
```diff
@@ -281,27 +286,36 @@ protected function constructPayload($sessions)
...
         try {
-            $http->post('', [
-                'json' => $payload,
-                'headers' => $headers,
-            ]);
+            if (method_exists($http, 'request')) {
+                $http->request('POST', '', $options);
+            } else {
+                $http->post('', $options);
+            }
```
imjoehaines, line 318: This isn't ideal but the return type of `Configuration::getSessionClient` is `GuzzleHttp\ClientInterface`, which doesn't have a `post` method in version 6 (it's defined on the concrete class instead)

We can tidy this up by depending on the concrete `GuzzleHttp\Client` class instead, but we have similar checks to this elsewhere so I've not done that as part of this PR

src/SessionTracker.php
imjoehaines: Actually we can't support `Countable` — we call `array_key_exists` elsewhere so we explicitly only support arrays here

GrahamCampbell: It would help to add phpdoc with type information to all the methods in this class, so the return type of `getSessionCounts` is clear. That function should probably throw away anything that is not null or an array, already.

src/SessionTracker.php
```diff
@@ -281,27 +290,36 @@ protected function constructPayload($sessions)
...
         ];
+
         $this->setLastSent();
 
         try {
-            $http->post('', [
-                'json' => $payload,
-                'headers' => $headers,
-            ]);
+            if (method_exists($http, 'request')) {
```
GrahamCampbell, line 318: This code should match https://github.com/bugsnag/bugsnag-php/commit/3264e0da5c4082337659870752cffa16a4e5920e#diff-add00846710b59ac0af2a07e1ede2747R209-R225. `method_exists` is technically wrong because it could detect something which is not a public method.

imjoehaines: Just making sure I understand — you mean check the `ClientInterface` specifically, rather than the `$http` object?

src/SessionTracker.php
imjoehaines: I've made this change in https://github.com/bugsnag/bugsnag-php/pull/579/commits/d2daba09ecd4d881a16e10291ae5304e57751153

I think it's safe because it's essentially the same as what we were doing internally anyway, but `getCurrentSession` is public so it could be a BC break. I'm not convinced it is though, given the bug we're fixing here is caused by returning the wrong type from the custom functions, so if anyone was relying on that it's likely they were triggering this bug (or a similar one with `lastSent`) anyway

src/SessionTracker.php
GrahamCampbell: Yes. This is because the interface tells you about the public API, rather than not necessarily public class methods.

PR #581, opened 2020-05-28
Async requests

GrahamCampbell:
Make http requests aync. I have defaulted this to be off in the http client constructor.

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -23,11 +23,25 @@ class HttpClient
     protected $guzzle;
 
     /**
+     * If async requests should be performed if possible.
+     *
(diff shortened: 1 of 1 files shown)


xPaw:
Isn't this the same thing as existing batch sending? (isBatchSending)

GrahamCampbell:
No. The difference between the two matters for long running processes.

imjoehaines:
I'm going to close this as this triggers an infinite loop in Guzzle (see https://github.com/guzzle/guzzle/issues/2752). We may be able to implement this in a future major version, where we have more leeway to refactor code to avoid causing the infinite loop

PR #582, opened 2020-06-04
Use HttpClient in SessionTracker

### LINKED ISSUE #558: Session handler design
Calling `makeGuzzle` is expensive, and should only be done once. Can we refactor the session config to not call it again?

imjoehaines:
## Goal

This change unifies the SessionTracker and Client WRT how they use Guzzle. They now both use the HttpClient abstraction, which also allows them to share a Guzzle instance. This means that we can't use the base URI feature of Guzzle anymore because the notify and session endpoints are on separate subdomains. However we always POST to the root so weren't really taking advantage of this anyway

### BC breaks

- Removal of `getSessionClient` method in `Client` and `Configuration`
  This doesn't make sense to keep given the session client is now the same as the notify Guzzle client. Keeping this would either mean changes to the "session" client also propagate to the notify client or changes to this do nothing. Either of which could lead to things being delivered to the wrong endpoint and expected Guzzle changes not being reflected in the actual requests

- We no longer use the Guzzle base URI for requests
  This means if a Guzzle client is passed in the constructor with a pre-defined base URI, we will still send requests to the notify URI. `Client::setNotifyEndpoint` now needs to be called manually instead. This changed because the Guzzle base URI is ambiguous now given that it is shared between notifications and sessions

- Removal of `SessionTracker::setConfig`
  This was unused internally and doesn't seem to be needed as the configuration can be changed via the `Client` or `Configuration` instance itself. Having the possibility for there to be two different `Configuration` instances could be dangerous as changes may not propagate as expected

- `Client::ENDPOINT`
  This is ambiguous as we have three separate endpoints
  Use `Configuration::NOTIFY_ENDPOINT` instead

- `HttpClient::PAYLOAD_VERSION`
  This is ambiguous as there is a session payload version too
  Use `HttpClient::NOTIFY_PAYLOAD_VERSION` instead

- `Report::PAYLOAD_VERSION`
  As above. This was also unused by the notifier
  Use `HttpClient::NOTIFY_PAYLOAD_VERSION` instead

- `SessionTracker::$SESSION_PAYLOAD_VERSION`
  Use `HttpClient::SESSION_PAYLOAD_VERSION` instead

- `Client::makeGuzzle` has been made private
  Use the `$guzzle` parameter of `Bugsnag\Client::__construct` instead

## Tests

The existing tests have been modified to avoid using the removed methods and the `SessionTracker` mocks updated to use `HttpClient` rather than Guzzle. Otherwise this is covered by existing tests

### Outstanding Questions

In the original PR for this (https://github.com/bugsnag/bugsnag-php/pull/521) we pulled the base URI out of the Guzzle instance that can be passed to the `Client` constructor and set the notify endpoint based on that

I'm not convinced whether that's a good idea — we can't use the base URI feature of Guzzle as we're sharing a single Guzzle client for two different URIs so they have to be absolute. That means if a base URI is set then it will be ignored anyway, because the absolute URI makes it irrelevant (the idea is you set `base_uri => 'https://example.com` and then can request `https://example.com/foo` by passing `foo` to Guzzle)

### Linked issues

Supersedes #521 
Fixes #558

## Review

  - [x] Final review


src/HttpClient.php
```diff
@@ -90,7 +136,10 @@ public function deploy(array $data)
 
         $data['apiKey'] = $this->config->getApiKey();
 
-        $this->post('deploy', ['json' => $data]);
+        $this->post(
+            $this->config->getNotifyEndpoint().'deploy',
```
imjoehaines, line 140: This change is quite fragile as it relies on the notify endpoint ending with a `/`, which we don't guarantee

I'm not sure this method is worth keeping though — it's been deprecated since https://github.com/bugsnag/bugsnag-php/pull/454 more than two years ago. We don't call this internally and the `HttpClient` isn't exposed anywhere. That means the only way to call this `deploy` method is by reflection trickery or extending `Client` and calling it from a custom implementation

Having said that, adding a `/` if one doesn't exist isn't difficult so we can also keep this around if that's preferable

src/Client.php
```diff
@@ -132,17 +143,15 @@ public function __construct(Configuration $config, ResolverInterface $resolver =
...
      *
-     * @return \GuzzleHttp\ClientInterface
+     * @deprecated Will be removed in a future release
      */
     public static function makeGuzzle($base = null, array $options = [])
     {
-        $key = method_exists(ClientInterface::class, 'request') ? 'base_uri' : 'base_url';
-
-        $options[$key] = $base ?: static::ENDPOINT;
-
```
tomlongridge, line 145: Shouldn't this be maintained until we remove the deprecation?

src/Configuration.php
```diff
@@ -590,84 +572,98 @@ public function shouldIgnoreErrorCode($code)
...
-    public function shouldCaptureSessions()
+    public function getBuildEndpoint()
     {
-        return $this->autoCaptureSessions;
+        return $this->buildEndpoint;
     }
 
     /**
-     * Sets the build endpoint.
+     * Set session tracking state and pass in optional guzzle.
```
tomlongridge, line 647:
```suggestion
     * Set session tracking state.
```

src/HttpClient.php
```diff
@@ -9,47 +9,43 @@
...
+    const PAYLOAD_VERSION = self::NOTIFICATION_PAYLOAD_VERSION;
 
     /**
-     * The queue of reports to send.
-     *
-     * @var \Bugsnag\Report[]
+     * The payload version for the error notification API.
      */
-    protected $queue = [];
+    const NOTIFICATION_PAYLOAD_VERSION = '4.0';
```
tomlongridge, line 24:
```suggestion
    const NOTIFY_PAYLOAD_VERSION = '4.0';
```

src/HttpClient.php
tomlongridge: I agree, we should look to remove if this ends up in a major release.

However, let's stick on a trivial check for a slash for now just so we maintain robust BC for now.

PR #584, opened 2020-06-09
Add caching abstraction around PSR-6, PSR-16 and Doctrine Cache

imjoehaines:
## Goal

This PR adds a caching abstraction which can wrap a PSR-6, PSR-16 or Doctrine Cache instance in a unified interface. More abstractions are easy to add by adding a wrapper class and a condition to the `CacheFactory`

This isn't used yet, but will be in an upcoming PR

## Tests

As PSR-6 and PSR-16 are only interfaces, I've added bare-minimum implementations so that we can write tests for them. These should eventually be expanded upon with integration tests against, e.g. a real Memcache instance, but that would require adding even more dev dependencies which I'd like to avoid. We can look into adding tests against an example project in another PR

The implementations are forced to behave the same via the `AdapterTest` abstract test case, which runs the same tests against all three implementations

## Review

  - [x] Final review

### DIFF
--- a/src/Cache/Adapter/DoctrineCacheAdapter.php
+++ b/src/Cache/Adapter/DoctrineCacheAdapter.php
@@ -38,8 +38,8 @@ final class DoctrineCacheAdapter implements CacheAdapterInterface
             }
 
             return $value;
-        } catch (Throwable $e) {
         } catch (Exception $e) {
+        } catch (Throwable $e) {
         }
 
         return $default;
@@ -49,8 +49,8 @@ final class DoctrineCacheAdapter implements CacheAdapterInterface
     {
         try {
             return $this->cache->save($key, $value);
-        } catch (Throwable $e) {
         } catch (Exception $e) {
+        } catch (Throwable $e) {
         }
 
         return false;
--- a/src/Cache/Adapter/DoctrineCacheAdapter.php
+++ b/src/Cache/Adapter/DoctrineCacheAdapter.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Bugsnag\Cache\Adapter;
+
+use Doctrine\Common\Cache\Cache as DoctrineCache;
+
+final class DoctrineCacheAdapter implements CacheAdapterInterface
+{
+    /**
+     * @var DoctrineCache
+     */
+    private $cache;
+
+    public function __construct(DoctrineCache $cache)
+    {
+        $this->cache = $cache;
+    }
+
+    public function get($key, $default = null)
+    {
+        $value = $this->cache->fetch($key);
+
+        if ($value === false) {
+            // Doctrine Cache specifies returning false if a key is not found.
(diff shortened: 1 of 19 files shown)


src/Cache/Adapter/DoctrineCacheAdapter.php
```diff
@@ -0,0 +1,74 @@
...
+            if ($value === false) {
+                // Doctrine Cache specifies returning false if a key is not
+                // found. However, 'false' is a possible value to cache so we
+                // need to make sure the key doesn't exist
+                // TODO we probably don't need to store 'false' in the cache so
+                //      can probably get away without doing this — this would
+                //      save a second round trip to the cache
+                if ($this->cache->contains($key)) {
+                    return false;
+                }
```
imjoehaines, line 35: I'm 99.9% sure we don't need this, but I figured I'd leave it until we're certain — happy to remove it now though

GrahamCampbell:
Note that some versions of Laravel don't implement the cache interface correctly. ;)

GrahamCampbell:
https://github.com/GrahamCampbell/Laravel-GitHub/blob/8.8/src/Http/Psr16Cache.php is a PSR-16 adapter for Laravel 5.5+.

imjoehaines:
> Note that some (older) versions of Laravel don't implement the cache interface correctly. ;)

Yikes, looks like it's just the TTL that would affect us though, which isn't super important because we should be using the cached values and resetting them way before hitting the TTL. I'm tempted to leave this as-is and wrap the cache (if necessary) in bugsnag-laravel

tests/TestCase.php
```diff
@@ -28,6 +28,24 @@ public function expectedException($class, $message = null)
         }
     }
 
+    protected function isPhpUnit7()
```
GrahamCampbell, line 31: this can be static

tests/TestCase.php
```diff
@@ -28,6 +28,24 @@ public function expectedException($class, $message = null)
...
+    {
+        return version_compare($this->phpUnitVersion(), '7.0.0', '>=')
+            && version_compare($this->phpUnitVersion(), '8.0.0', '<');
+    }
+
+    protected function isPhpUnit6()
```
GrahamCampbell, line 37: etc

tests/TestCase.php
```diff
@@ -36,18 +54,14 @@ protected static function getGuzzleMethod()
         return method_exists(ClientInterface::class, 'request') ? 'request' : 'post';
     }
 
-    private static function readObjectAttribute($object, $attributeName)
+    private function phpUnitVersion()
```
GrahamCampbell, line 57: etc

tests/TestCase.php
imjoehaines: It can be; I don't think there's a compelling reason for this to be static though?

GrahamCampbell: Normally one would write it static unless there is a compelling reason not to.

PR #587, opened 2020-06-17
Use one Guzzle instance for everything

imjoehaines:
## Goal

This change unifies the SessionTracker and Client WRT how they use Guzzle. They now both use the HttpClient abstraction, which also allows them to share a Guzzle instance, saving the cost of instantiating Guzzle multiple times and allowing configuration to be shared

This means that we can't use the base URI feature of Guzzle anymore because the notify and session endpoints are on separate subdomains. However we always POST to the root so weren't really taking advantage of this anyway

I've also fixed a bug with `trimOldestSessions` where it was trimming random sessions because of an unstable sorting function. Now it works as expected and has a test too

This PR builds on https://github.com/bugsnag/bugsnag-php/pull/582, which made similar changes but with BC breaks as we've decided to avoid a major version bump. See also #521

The deprecations are listed in the changelog, but the exact wording needs some work — I'd like to get the PR reviewed first. The majority of these are also quite unlikely to affect many people as they relate to internal constants or protected methods (which are technically part of the public API, but only if you extend the class)

## Tests

I've added a bunch of tests to the `SessionTracker` and change the tests around the `Client` endpoints

Most tests are essentially the same as before, so BC should be preserved (at least for the things that are tested)

### Linked issues

Related to #582 
Related to #521


composer.json
```diff
@@ -21,7 +21,8 @@
         "mockery/mockery": "^0.9.4|^1.3.1",
         "mtdowling/burgomaster": "dev-master#72151eddf5f0cf101502b94bf5031f9c53501a04",
         "phpunit/phpunit": "^4.8.36|^7.5.15",
-        "php-mock/php-mock-phpunit": "^1.1|^2.1"
+        "php-mock/php-mock-phpunit": "^1.1|^2.1",
+        "sebastian/version": ">=1.0.3"
```
GrahamCampbell, line 25: Missing upper bound.

src/Client.php
```diff
@@ -135,13 +157,13 @@ public function __construct(Configuration $config, ResolverInterface $resolver =
      * @param string|null $base
      * @param array       $options
      *
-     * @return \GuzzleHttp\ClientInterface
```
GrahamCampbell, line 138: Using fully qualified names in phpdoc makes it human readable as well as machine readable.

composer.json
imjoehaines: I'm not sure if we need an upper bound here — we don't use this directly but go via PHPUnit's `PHPUnit\Runner\Version` class so BC breaks shouldn't affect us unless they're also BC breaks in PHPUnit, which we'd need to fix anyway. It also doesn't really change; the only new release in the last 4 years was essentially a PHP version bump

We just need a lower bound when using `--prefer-lowest` to avoid [a bug in v1.0.0-1.0.2](https://github.com/sebastianbergmann/version/issues/7) where it reads our version number rather than PHPUnit's

src/Client.php
imjoehaines: I don't really agree; it makes the code harder to read and the overwhelming convention in the community is to use the imported names

Having said that, the `Client` <-> `ClientInterface` could be confusing; solved in 664e9de20ea2142d875427e23e4fa02496e43cf9

src/Client.php
```diff
@@ -78,19 +79,28 @@ class Client
...
         $env = new Env();
 
         $config = new Configuration($apiKey ?: $env->get('BUGSNAG_API_KEY'));
-        $guzzle = static::makeGuzzle($endpoint ?: $env->get('BUGSNAG_ENDPOINT'));
+
+        $notifyEndpoint = $notifyEndpoint ?: $env->get('BUGSNAG_ENDPOINT');
+
+        if (is_string($notifyEndpoint)) {
+            $config->setNotifyEndpoint($notifyEndpoint);
+        }
```
tomlongridge, line 101: Duplicate code in constructor?

src/Client.php
```diff
@@ -102,23 +112,34 @@ public static function make($apiKey = null, $endpoint = null, $defaults = true)
...
+        $guzzle = $guzzle ?: self::makeGuzzle();
+
+        // We need to use 'getBaseUrl' for Guzzle v5 compatibility
+        $base = method_exists(GuzzleHttp\ClientInterface::class, 'getBaseUrl')
+            ? $guzzle->getBaseUrl()
+            : $guzzle->getConfig(self::getGuzzleBaseOptionName());
+
+        if (is_string($base) || method_exists($base, '__toString')) {
+            $config->setNotifyEndpoint((string) $base);
+        }
```
tomlongridge, line 135: Consider abstracting Guzzle version logic as per `getGuzzleBaseOptionName`

src/Configuration.php
```diff
@@ -7,22 +7,21 @@
 class Configuration
 {
     /**
-     * The default endpoint.
-     *
-     * @var string
+     * The default endpoint for error notifications.
```
tomlongridge, line 10:
```suggestion
     * The default endpoint for event notifications.
```

src/Configuration.php
```diff
@@ -590,84 +592,114 @@ public function shouldIgnoreErrorCode($code)
     }
 
     /**
-     * Set session tracking state and pass in optional guzzle.
+     * Set notification delivery endpoint.
```
tomlongridge, line 595:
```suggestion
     * Set event notification endpoint.
```

src/Configuration.php
```diff
@@ -590,84 +592,114 @@ public function shouldIgnoreErrorCode($code)
...
+    public function setNotifyEndpoint($endpoint)
     {
-        $this->autoCaptureSessions = $track;
+        $this->notifyEndpoint = $endpoint;
 
         return $this;
     }
 
+    /**
+     * Get notification delivery endpoint.
```
tomlongridge, line 609:
```suggestion
     * Get event notification endpoint.
```

CHANGELOG.md
```diff
@@ -1,6 +1,58 @@
...
+  Use `HttpClient::getEventPayload` instead
+
+- `HttpClient::postJson`
+  Use `HttpClient::deliverEvents` instead
+
+- Using the `base_uri`/`base_url` on a Guzzle instance
+  The base URI is ambiguous as there are three separate endpoints which could be used, therefore all Guzzle requests now use absolute URIs. We will extract the `base_uri`/`base_url` Guzzle option if one is set and use it as the notification endpoint URI, however this will be removed in the next major version
+  Set the notification endpoint manually with `Configuration::setNotifyEndpoint` instead
+
+- Calling `HttpClient::getHeaders` without providing a payload version
```
tomlongridge, line 40: Missing deprecation notice?

src/HttpClient.php
```diff
@@ -224,14 +272,29 @@ protected function post($uri, array $options = [])
...
+
+    /**
+     * Deliver the given events to the notification API.
+     *
+     * @param string $uri  the uri to hit
+     * @param array  $data the data send
+     *
+     * @return void
+     */
+    protected function deliverEvents($uri, array $data)
```
tomlongridge, line 297: The naming of the event delivery functions is a little different to sessions and builds. Once we remove the deprecations, we'll have `sendBuildReport(array $data)`, `sendSession(array $data)` but `sendEvents()` and `deliverEvents($uri, array $data)`.

As we're making deprecations, I think we should aim towards `sendEvents(array $data)`. This may mean you need a lower-level private function that takes an array and URI which you can use for `postJson`. Then switch naming around so that `deliverEvents` is the parameterless method that reads the queue and is called from `send` until that is deprecated.

src/Client.php
GrahamCampbell: The community only do this because they have IDEs written by Java programmers. Laravel, for example, uses fully qualified class names, only.

GrahamCampbell: Also, importing a class only to use it in the phpdoc actually has a runtime performance hit in PHP, unlike in Java.

composer.json
GrahamCampbell: `^1.0.3` then?

src/HttpClient.php
imjoehaines: Are you saying to move the queued events somewhere else? I don't think it makes sense to make `sendEvents` take the array of events, unless `HttpClient` doesn't track the queue of events itself — otherwise the thing calling `sendEvents` has to read the queue from `HttpClient` first, then pass that back to `HttpClient` to send

tomlongridge: It's just a naming thing - we have `sendXXX` elsewhere that takes a param with what to send. Whereas for events it's called `deliverEvents` that takes the parameter and `sendEvents` reads the queue. Was just suggesting to reverse the naming.

imjoehaines: Wouldn't that make the public interface of `HttpClient` `sendSessions`, `sendBuildReport` and `deliverEvents` though?

I think it'd be nice to move the queue elsewhere and make the `sendX` methods all take some data to send, but maybe in a separate PR?

PR #594, opened 2020-09-03
Improve automated tests for Bugsnag\Handler

imjoehaines:
## Goal

Improve the automated tests for `Bugsnag\Handler`, in preparation for some changes in behaviour. These tests currently surface some issues with the handler, which will be fixed/changed separately

It's probably easier to review each commit separately, rather than everything at once

## Changeset

- Added PHPT tests for `Bugsnag\Handler`
- Added fake Guzzle implementations for use in PHPT tests, which print when a request is made

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -9,3 +9,4 @@ enabled:
 finder:
   exclude: Bugsnag
   not-name: catchall.php
+  not-path: tests/phpt/fixtures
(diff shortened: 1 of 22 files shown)


.styleci.yml
```diff
@@ -9,3 +9,4 @@ enabled:
 finder:
   exclude: Bugsnag
   not-name: catchall.php
+  not-path: tests/phpt/fixtures
```
imjoehaines, line 12: Style CI fails if it tries to do analysis on the deliberate parse error here

kattrali (approved):

PR #596, opened 2020-09-04
Filter cookies from events by default

imjoehaines:
## Goal

In other notifiers, the cookies header is filtered by default and we don't collect a 'Cookies' tab — this PR brings PHP inline

Cookies can still be recorded by a user by removing it as a filtered value and adding the cookie callback. The filter controls whether it shows in 'Request &rarr; Headers' and the callback adds the 'Cookies' tab

```php
$bugsnag = \Bugsnag\Client::make();

$bugsnag->setFilters(['password']);
$bugsnag->registerCallback(
    new \Bugsnag\Callbacks\RequestCookies(
        new \Bugsnag\Request\BasicResolver()
    )
);
```

## Changeset

- No longer register the `RequestCookies` callback by default
- `Configuration::$filters` now contains `cookie`, to filter the cookie header
- Added a test for the default callbacks, which also ensures cookies aren't collected as a separate tab and the cookie header is filtered

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -418,6 +418,17 @@ class ClientTest extends TestCase
                 $this->assertSame($expectedMetadata, $report->getMetaData());
                 $this->assertSame(['id' => '8.76.54.321'], $report->getUser());
                 $this->assertSame('GET /abc/xyz', $report->getContext());
+
+                $payload = $report->toArray();
+
+                $this->assertSame(
+                    [
+                        'Host' => 'example.com',
+                        'Cookie' => '[FILTERED]',
+                        'X-Forwarded-For' => '8.76.54.321',
+                    ],
+                    $payload['metaData']['request']['headers']
+                );
             }
         );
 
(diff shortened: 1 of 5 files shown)


tests/ClientTest.php
```diff
@@ -362,6 +364,77 @@ public function testNotOverriddenByCallbacks()
...
+    {
+        $_SERVER['REQUEST_METHOD'] = 'GET';
+        $_SERVER['REMOTE_ADDR'] = '123.45.67.8';
+        $_SERVER['HTTP_HOST'] = 'example.com';
+        $_SERVER['HTTP_COOKIE'] = 'tastes=delicious';
+        $_SERVER['HTTP_X_FORWARDED_FOR'] = '8.76.54.321';
+        $_SERVER['REQUEST_URI'] = '/abc/xyz?abc=1&xyz=2';
+        $_GET['abc'] = '1';
+        $_GET['xyz'] = '2';
+        $_COOKIE['tastes'] = 'delicious';
```
kattrali, line 377: 😆

kattrali (approved):

PR #600, opened 2020-09-14
Error handler improvements

imjoehaines:
## Goal

This PR adds a number of improvements to the error handler:

1. Report new exceptions raised by the previous error handler
2. Allow the default PHP exception handler to run if the previous handler wants it to (by re-raising the exception that's being handled)
3. Avoid double reporting exceptions if the previous handler re-raises them (currently this will report once in the exception handler and once in the shutdown handler)

Because of these improvements we've also decided to always call the previous handlers, regardless of if `register` or `registerWithPrevious` is used

It's still possible to prevent Bugsnag from calling the previous handlers like this:

```php
// Assuming $client is an instance of \Bugsnag\Client
$handler = new \Bugsnag\Handler($client);

$handler->registerErrorHandler(false);
$handler->registerExceptionHandler(false);
$handler->registerShutdownHandler();
```

Or the "chain" of error/exception handlers can be broken by creating handlers that do nothing:

```php
set_error_handler(function () {});
set_exception_handler(function () {});

\Bugsnag\Handler::register();
```

## Design

Some of the changes are a bit unintuitive, but the PHPT tests should show why they are necessary. The technique of re-throwing the exception inside the exception handler especially, but it's important to solve issues like #523 — this PR doesn't entirely fix that issue as we also need to re-raise exceptions when there is no previous handler, which will be done separately

## Testing

Mostly tested through the PHPT tests and manually

The unit test has been re-written to use much less global mocking by setting/restoring error/exception handlers in the tests (one test still has a mocked global function). Unfortunately git tried too hard to diff the old & new file, so the diff is useless for reviewing ☹️

### DIFF
--- a/tests/phpt/handler_should_call_the_previous_exception_handler.phpt
+++ b/tests/phpt/handler_should_call_the_previous_exception_handler.phpt
@@ -8,7 +8,7 @@ set_exception_handler(function ($throwable) {
     var_dump($throwable);
 });
 
-Bugsnag\Handler::registerWithPrevious($client);
+Bugsnag\Handler::register($client);
 
 throw new RuntimeException('abc xyz');
 
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -63,6 +63,8 @@ class Handler
      * @param \Bugsnag\Client|string|null $client client instance or api key
      *
      * @return static
+     *
+     * @deprecated Use {@see Handler::register} instead.
      */
     public static function registerWithPrevious($client = null)
     {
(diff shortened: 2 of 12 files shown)


tests/phpt/handler_should_call_the_previous_exception_handler.phpt
```diff
@@ -1,17 +1,14 @@
 --TEST--
 Bugsnag\Handler should call the previous exception handler
-
-TODO this is currently reported twice! Once in the exceptionhandler and once on shutdown
```
imjoehaines, line 4: The new `enableShutdownHandler` flag fixes this bug ✨

kattrali (approved):

PR #609, opened 2020-10-13
Fixed branch alias

GrahamCampbell:
The purpose of branch aliases is to indicate to Packagist what version the branch is. That is, what release series to expect the next tag on the master branch to be. This allows people to test with `@dev` stability, and get the correct version of Bugsnag.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.20-dev"
+            "dev-master": "3.23-dev"
         }
     },
     "minimum-stability": "dev",


imjoehaines:
Can you explain the use-case that needs this change?

We only ever push to master for a new release which gets immediately tagged, so I'm not sure if this is useful; the latest version of master is always the same as the latest tagged release

GrahamCampbell:
The purpose of this is so that composer knows what version the code on that branch will be. If someone asks composer to get the latest versions, not just the latest stable versions, then composer will think the code on this branch represents wip for the next 3.20.x release, which is wrong.

This feature is particularly useful when someone wants to test a minor release series before a first .0 tag has been made, or just wants to get bug fixes before they are tagged.

PR #612, opened 2020-10-26
Increase timestamp resolution to include milliseconds

imjoehaines:
## Goal

In order for the Bugsnag app to report breadcrumb times more accurately, we need to send timestamps with milliseconds. Other timestamps have also been updated for consistency, but shouldn't have any impact. The session tracker has not been updated as it intentionally rounds to the minute

This includes some "fun" workarounds in order to support PHP 5 as the millisecond format character (`v`) was introduced in 7.0. The final format we want has a constant which was also introduced in PHP 7.0 (`RFC3339_EXTENDED`), however I've implemented PHP 5 compatible formatting on every version, to avoid the possibility of bugs related to using PHP 5/7/8

## Testing

Unit tests now check the date format in places that have changed. I've manually tested that the breadcrumb timestamps appear correctly to the millisecond

### DIFF
--- a/tests/DateTime/ClockTest.php
+++ b/tests/DateTime/ClockTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Bugsnag\Tests\DateTime;
+
+use Bugsnag\DateTime\Clock;
+use Bugsnag\DateTime\ClockInterface;
+use Bugsnag\Tests\TestCase;
+use DateTimeImmutable;
+
+class ClockTest extends TestCase
+{
+    public function testItImplementsClockInterface()
+    {
+        $clock = new Clock();
+
+        $this->assertInstanceOf(ClockInterface::class, $clock);
+    }
+
+    public function testItReturnsADateTimeImmutable()
+    {
+        $clock = new Clock();
+
+        $this->assertInstanceOf(DateTimeImmutable::class, $clock->now());
+    }
+
+    public function testItReturnsTheCurrentDateTime()
+    {
+        // We need to sleep between creating date objects, otherwise we hit
+        // issues due to $now possibly being equal to $before
+
+        // Sleeping for 5ms is enough on PHP 7.1+
+        $timeToSleep = 5000;
+
+        // Before PHP 7.1, microseconds were not included when constructing date
+        // objects, so we need to sleep for an entire second
+        if (version_compare(PHP_VERSION, '7.1.0', '<')) {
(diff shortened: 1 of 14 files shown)


tests/DateTime/ClockTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function testItReturnsTheCurrentDateTime()
+    {
+        $clock = new Clock();
+
+        $before = new DateTimeImmutable();
+
+        $now = $clock->now();
+
+        $after = new DateTimeImmutable();
```
kattrali, line 34: Is this actually getting values aside from all three being equal? Could pop a `usleep` in here to avoid the equal cases.

kattrali (approved):

tests/DateTime/ClockTest.php
imjoehaines: They _should_ all be distinct because `new DateTimeImmutable` includes microseconds, so it's unlikely they'll be the exact same but I let it happen to avoid flakes. Locally they are never the same but maybe with a super PC you could make all 3 dates in the same microsecond :p

I guess a super short `usleep` should be enough to ensure it doesn't ever flake and also not have any noticeable impact on the test length

The other test _needs_ to `usleep` because `Date::now()` only has milliseconds, so technically is before `$before` e.g.

```php
$before = new DateTimeImmutable();
$now = new DateTimeImmutable(Bugsnag\DateTime\Date::now());

var_dump($before, $now);
var_dump($before < $now);
```

```
object(DateTimeImmutable)#3 (3) {
  ["date"]=>
  string(26) "2020-10-26 13:56:23.649783"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(3) "UTC"
}
object(DateTimeImmutable)#2 (3) {
  ["date"]=>
  string(26) "2020-10-26 13:56:23.649000"
  ["timezone_type"]=>
  int(1)
  ["timezone"]=>
  string(6) "+00:00"
}
bool(false)
```

We could construct `$before` without microseconds, but we'd either still have to `usleep` or allow them to be equal

kattrali: Ah! That was actually more complex than I realized 🤔 💭

kattrali (approved):

tests/DateTime/ClockTest.php
imjoehaines: PHP! 🐘

PR #618, opened 2020-11-27
Remove 'ignore-platform-reqs' from PHP 8 CI

imjoehaines:
## Goal

Remove the `ignore-platform-reqs` composer flag from PHP 8 CI. This was necessary to install dependencies on PHP 8 before they officially supported them, but is not necessary anymore


kattrali (approved):
took a long time to review but lgtm. 🙃

PR #620, opened 2021-01-26
Disable XDebug on CI and register default callbacks in PHPT tests

imjoehaines:
## Goal

We're currently running CI with XDebug which has performance implications and also makes our PHPT tests fail because it emits a warning to stdout:
> Xdebug: [Config] The setting 'xdebug.default_enable' has been renamed, see the upgrading guide at https://xdebug.org/docs/upgrade_guide#changed-xdebug.default_enable (See: https://xdebug.org/docs/errors#CFG-C-CHANGED)

We don't use XDebug for anything so can safely disable it

Our PHPT tests also don't register the default callbacks, which are important as they attach metadata to events

### DIFF
--- a/tests/phpt/handler_can_be_manually_called_multiple_times.phpt
+++ b/tests/phpt/handler_can_be_manually_called_multiple_times.phpt
@@ -17,7 +17,7 @@ $handler->exceptionHandler(new LogicException('terrible things'));
 --EXPECTF--
 array(1) {
   [0]=>
-  object(Exception)#15 (7) {
+  object(Exception)#%d (7) {
     ["message":protected]=>
     string(10) "bad things"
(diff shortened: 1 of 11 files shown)


tests/phpt/handler_can_be_manually_called_multiple_times.phpt
```diff
@@ -17,7 +17,7 @@ $handler->exceptionHandler(new LogicException('terrible things'));
 --EXPECTF--
 array(1) {
   [0]=>
-  object(Exception)#15 (7) {
+  object(Exception)#%d (7) {
```
imjoehaines, line 20: These object IDs changed because we now instantiate the default callback classes — they should have always been "any number"

kattrali (approved):

PR #621, opened 2021-01-28
Allow Bugsnag\Handler to handle OOMs

imjoehaines:
## Goal

Currently we are able to report OOMs that are caused by a single, large allocation as PHP doesn't actually allocate the memory so there's plenty left for us to use. However, when an OOM happens due to lots of small allocations, we crash as there's not enough memory left to send the report

## Design

`Bugsnag\Handler` will now increase the memory limit on an OOM by the number of bytes specified in the new `memoryLimitIncrease` option (5MiB default)

To do this successfully, we need to reserve a bit of memory so that we can:

1. check if an error happened using `error_get_last`
1. check if it was an OOM
 
We don't reserve all of the memory that we need upfront, as this would be a big overhead on every request when it shouldn't be necessary most of the time. 32K was chosen because `error_get_last` can cause PHP to re-allocate chunks of memory that it has already allocated. This can lead to a second OOM if the chunks it attempts to re-allocate are larger than the available memory. 32K seems to be a happy medium between being small enough to not matter and large enough to allow `error_get_last` to be called, but this can be tweaked in future

## Changeset

- Added a new configuration option `memoryLimitIncrease` (5MiB default), with getters & setters in `Client` & `Configuration`
- `Handler` reserves 32K of memory when the shutdown function is registered
- `Handler` increases the current memory limit by `memoryLimitIncrease` if an OOM is detected

## Testing

New PHPT tests cause OOMs and check delivery

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Out of memory errors will now be reported by increasing the memory limit by 5 MiB. Use the new `memoryLimitIncrease` configuration option to change the amount of memory, or set it to `null` to disable the increase entirely.
+  [#621](https://github.com/bugsnag/bugsnag-php/pull/621)
+
 ## 3.25.0 (2020-11-25)
 
 ### Enhancements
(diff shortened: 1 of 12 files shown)


kattrali (approved):
nice work!

PR #623, opened 2021-02-05
Add new 'redacted keys' config option to replace 'filters'

imjoehaines:
## Goal

This PR adds a 'redacted keys' option, to replace the existing 'filters' option. Filters is problematic because it works with a case-insensitive string contains, which can easily cause false positives

Redacted keys has different semantics but can still express everything that filters could (and more!). Redacted keys can contain both strings and regexes, where string comparisons are case-insensitive and exact — i.e. `a` matches `a` and `A` but not `ab`. Regexes can be used to emulate filters, e.g. `/a/i` will match `a`, `A` and `ab`

Redacted keys also attempts to match multi-byte strings by normalising strings and using multi-byte comparisons. This relies on the intl and mbstring extensions, so we fallback to `strcasecmp` if they are missing (we also try `strcasecmp` first as it's quicker than normalising + MB comparisons)

## Changeset

- New config option `redactedKeys` with getters & setters in Client & Configuration
- Existing `filters` option is now marked as deprecated
- `Report::shouldFilter` also uses `redactedKeys` to check if keys should be filtered/redacted
- A new `Utils::stringCaseEquals` method has been added. This does a unicode-aware, case-insensitive string match
- Tests

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -29,6 +29,7 @@ jobs:
       with:
         php-version: ${{ matrix.php-version }}
         coverage: none
+        extensions: intl, mbstring
 
     - run: composer validate
 
--- a/tests/phpt/handler_should_increase_memory_limit_by_configured_amount_on_oom.phpt
+++ b/tests/phpt/handler_should_increase_memory_limit_by_configured_amount_on_oom.phpt
@@ -5,24 +5,24 @@ Bugsnag\Handler should increase the memory limit by the configured amount when a
 $client = require __DIR__ . '/_prelude.php';
 $client->setMemoryLimitIncrease(1024 * 1024 * 10);
 
-ini_set('memory_limit', '2M');
+Bugsnag\Handler::register($client);
+
+ini_set('memory_limit', 1024 * 1024 * 5);
 var_dump(ini_get('memory_limit'));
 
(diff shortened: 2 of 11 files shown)


tests/phpt/handler_should_increase_memory_limit_by_configured_amount_on_oom.phpt
```diff
@@ -5,24 +5,24 @@ Bugsnag\Handler should increase the memory limit by the configured amount when a
 $client = require __DIR__ . '/_prelude.php';
 $client->setMemoryLimitIncrease(1024 * 1024 * 10);
 
-ini_set('memory_limit', '2M');
+Bugsnag\Handler::register($client);
```
imjoehaines, line 8: This test could run OOM _before_ installing the handler, so I've moved this to before we reduce the memory limit and bumped the limit a bit

.github/workflows/test-package.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       with:
         php-version: ${{ matrix.php-version }}
         coverage: none
+        extensions: intl, mbstring
```
imjoehaines, line 32: PHPUnit requires mbstring so we can't run the tests without it ☹️

kattrali (approved):

PR #632, opened 2021-10-19
Support sending arguments with stacktraces

### LINKED ISSUE #519: Argument in stacktrace
### Description
I would like access arguments from the stacktrace inside the notification callback, which is currently not possible.

It's clear that this is a design choice atm given the following code:
` PHP
// Reduce memory usage by omitting args and objects from backtrace
$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS & ~DEBUG_BACKTRACE_PROVIDE_OBJECT);
`

It would be nice if there was a config option to change this. 

If you also see a benefit in this or would suggest a different solution,  I am happy to provide a PR.

GrahamCampbell:
## Goal

As a developer, I'd like to see the arguments passed to functions in stack trace frames to enable me to better understand what happened and how to reproduce the issue. Closes #519.

## Design

The `Configuration` class has a mechanism for toggling if code should be sent. We can copy this design to add support for toggling if argument values should be included. Objects, arrays and resources are represented by noting their type but not their value. Quite a rabbit hole otherwise...

## Changeset

New `setSendArguments` and `shouldSendArguments` functions. Disabled by default, so no change in functionality for existing users.

## Testing

Unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## 3.27.0 (UPCOMING)
+
+### Enhancements
+
+* Support sending arguments with stack traces, disabled by default.
+  [#632](https://github.com/bugsnag/bugsnag-php/pull/632)
+
 ## 3.26.1 (2021-09-09)
 
 ### Fixes
--- a/src/Client.php
+++ b/src/Client.php
@@ -656,6 +656,30 @@ class Client
     }
 
     /**
+     * Sets whether arguments should be captured in stack traces.
+     *
+     * @param bool $sendArguments whether to send arguments to Bugsnag
+     *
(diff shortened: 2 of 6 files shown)


mattdyoung:
Hi @GrahamCampbell 

Thanks for the PR. Unfortunately we've decided that we're not going to move forward with this feature for the time being.

There are a few wider issues that would need to be addressed such as dashboard updates to ensure long values wrap sensibly, and providing a new API with finer control over redacting data as the arguments are highly likely to contain sensitive data in some applications. We'd need to invest some effort in designing an implementing these changes to fully support this in an acceptable way.

We're continuing to monitor interest on the original github issue and may pick this up again in the future when priorities allow.

PR #635, opened 2021-12-07
Fix PHAR build

imjoehaines:
## Goal

The CA certs in the docker image have finally expired so anything using SSL stopped working

Updating the packages fixes this as it pulls in new versions of curl/libssl/ca-certificates and friends

I've also switched over to copying Composer from their docker image as it's a bit faster than downloading & installing it, especially once cached

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM php:5.5.38-alpine
+FROM php:5.5-alpine
 
 ENV COMPOSER_ALLOW_SUPERUSER 1
 
-RUN curl -sS https://getcomposer.org/installer | php \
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
You could build the phar on a PHP 7.4 image, but tell composer to emulate PHP 5.5 when resolving dependencies?

imjoehaines:
> You could build the phar on a PHP 7.4 image, but tell composer to emulate PHP 5.5 when resolving dependencies?

I'd prefer to avoid big changes to this and I'm not sure if Burgomaster runs on PHP 7 anyway

GrahamCampbell (approved):

twometresteve (approved):

PR #633, opened 2021-11-18
Add phpstan and fix up types

xPaw:
There were plenty of incorrect types, I've fixed up most of them.

phpstan still reports various errors which should be fixed, but I'll leave that. If you're not really interested in phpstan, type fixes can be merged without it.

PR #637, opened 2022-01-31
Add PHPStan at level 6

imjoehaines:
## Goal

PHPStan is a static analyser for PHP that can [catch a bunch of potential bugs in code](https://phpstan.org/blog/find-bugs-in-your-code-without-writing-tests)

This PR is based on https://github.com/bugsnag/bugsnag-php/pull/633, but runs at level 6 instead of max — PHPStan has 9 error levels (currently) that go from loosest (1) - strictest (9)

For the most part this is only changes to comments, but there are a couple of code changes too:

- methods that can return `null` need an explicit `return null` statement — a bare `return` or no `return` at all [are both errors](https://phpstan.org/r/37d47edc-c51c-40b8-b32a-652994a843c0)
- passing an `int` to `ini_set` and passing `null` to `preg_match` are flagged as errors, so these have been fixed

Additionally, some errors have been ignored:

- some of our Guzzle compatibility code errors because we check for methods that don't exist on new versions. We'll need to run PHPStan against multiple versions of Guzzle to fully check these (this will be done in a future PR)
- some code exists for PHP 5 compatibility, but PHPStan doesn't run on PHP <7.1 so this errors as well. When we drop PHP 5 support we can also remove this compatibility code, which will solve these errors
- we use `new static(...)` in a few places, which is unsafe as there's guarantee that child classes will share the same constructor signature ([see this article](https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static)). We can't fix this without a potential BC break, so have to ignore this for now

## Design

This PR stops at level 6 because this is where the levels become tricky to update to. Level 6 specifically adds checking for iterable types, e.g. a bare `array` should really be `array<KeyType, ValueType>`. I've turned this option off for now as we get around 100 errors from it, but this will be turned on in a future PR. Until then, I've not advanced any further as additional errors can appear as a consequence of adding stricter iterable types

## Testing

PHPStan now runs on CI on both PHP 7.1 & 8.1

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -49,3 +49,42 @@ jobs:
       run: composer update --prefer-dist --no-progress --no-suggest --no-interaction ${{ matrix.composer-flags }}
 
     - run: composer run-script test
+
+  phpstan:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # TODO: we likely have to care about Guzzle's version too - ideally we'd
+        #       check our use of Guzzle is correct on each version
+        php-version: ['7.1', '8.1']
+        phpstan-version: ['~1.4']
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: install PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-version }}
+        coverage: none
+        extensions: intl, mbstring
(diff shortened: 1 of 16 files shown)


xPaw:
FYI, once PHP5 support is dropped, [types can be added to function arguments](https://www.php.net/manual/en/language.types.declarations.php)

kattrali (approved):

PR #639, opened 2022-02-02
Support enums in metadata

imjoehaines:
## Goal

This PR improves our handling of enums when added as metadata

PHP has two types of enums, "pure" and "backed". A pure enum contains only cases with no values, whereas a backed enum has a string or int value for each case. For example:

```php
enum PureEnum {
    case One;
    case Two;
}

enum StringBackedEnum: string {
    case One = 'one';
    case Two = 'two';
}

enum IntBackedEnum: int {
    case One = 1;
    case Two = 2;
}
```

Currently when a pure enum is added as metadata it falls into our generic object handling, which ends up converting it to `null` as [there is no default serialisation for pure enums](https://3v4l.org/CinD6#v8.1.2)

This PR fixes this by converting the enum to a string in the form `EnumName::CaseName`, e.g. `PureEnum::One` using the example above

Backed enums aren't broken as they are converted into their backing values, e.g. `StringBackedEnum::One` is serialised as the string `one`. This could be improved however, given the enum name is usually more important than the raw value

This PR changes backed enums to be serialised in the form `EnumName::CaseName (value)`, e.g. `IntBackedEnum::One (1)`

These changes result in the following being shown in the Bugsnag dashboard:

image

## Testing

Two new PHPT tests have been added; one that tests pure enums and one that tests backed enums. PHPT tests are being used because PHP versions without enum support will fail to compile files that declare enums, making it a bit awkward to unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Changelog
   [xPaw](https://github.com/xPaw)
   [#633](https://github.com/bugsnag/bugsnag-php/pull/633)
   [#637](https://github.com/bugsnag/bugsnag-php/pull/637)
+* Handle serialising pure enums when added as metadata. Previously a pure enum would be JSON encoded as `null`, but will now be converted to a string like `EnumName::CaseName`
+  [#639](https://github.com/bugsnag/bugsnag-php/pull/639)
 
 ## 3.26.1 (2021-09-09)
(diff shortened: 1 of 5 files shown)


kattrali (approved):
looks good! still envious of phpt tests for other platforms. lol

PR #653, opened 2022-10-18
Add `Report::getFeatureFlags` to allow accessing feature flags in callbacks

imjoehaines:
Add `Report::getFeatureFlags` to allow accessing feature flags in callbacks. This returns a list of `FeatureFlag` instances (the same class used in `addFeatureFlags`)

I've changed the return value of `FeatureFlagDelegate::toArray` to the raw array of feature flags to support this. The `Report::toArray` now handles formatting the list of feature flags for the Event API


kattrali (approved):
nice!

PR #672, opened 2024-12-06
PHP 8.4 support

agrzegorzewski:
## Goal

With the recent release of PHP 8.4, some of the features that were used by the notifier were marked as deprecated and started throwing depreciation warnings. This PR removes both of the depreciation warnings without breaking backwards compatibility.

https://php.watch/versions/8.4/E_STRICT-deprecated
https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated
 
## Design

The issue with nullable types was solved by removing type annotations from parameters that default to `null`. Type annotations provided in the doc comments above already provide accurate annotations, so LSPs should continue to infer types correctly.

The depreciation of `E_STRICT` required more widespread changes. Due to conditional properties in associative arrays not being supported in <6.0, the list of errors had to be rewritten with an initializer that conditionally adds the value to the array. This also warranted a change in tests that rely on `E_STRICT`.

The added tests for PHP 8.3 and 8.4 required small changes in the unit tests to account for differences in error messages.

## Changeset

- Removed type hints from optional parameters.
- Removed the usage of `E_STRICT` in PHP 8.4+.
- Added PHP 8.3 and 8.4 to tests.
- Adjusted unit tests.

## Testing

Tested manually and automatically. No new tests were added. Some of the existing tests were extended.

### DIFF
--- a/tests/phpt/handler_should_report_user_errors.phpt
+++ b/tests/phpt/handler_should_report_user_errors.phpt
@@ -15,7 +15,7 @@ var_dump('I should not be reached');
 --SKIPIF--
 <?php
 if (PHP_VERSION_ID >= 80400) {
-    echo 'SKIP — this test uses methods depreciated in PHP 8.4+';
+    echo 'SKIP — this test uses methods deprecated in PHP 8.4+';
 }
 ?>
 --EXPECTF--
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -23,6 +23,10 @@ jobs:
           guzzle-version: '^7.0'
         - php-version: '8.2'
           guzzle-version: '^7.0'
+        - php-version: '8.3'
+          guzzle-version: '^7.0'
+        - php-version: '8.4'
+          guzzle-version: '^7.9'
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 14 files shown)


agrzegorzewski:
I have no clue why PHP 7.2 tests suddenly fail. It feels related to https://github.com/bugsnag/bugsnag-laravel/pull/544

agrzegorzewski:
I tried applying the same fix as in that Laravel PR, but it didn't help

Cawllec (commented):
I've tested this against our previous versions, and it seems to resolve the 8.4 issues nicely!  I've also tested with PHP 5, and can confirm it works there as well.

Could we retarget the merge into `next`, then I can take care of the changelog and version bumps, and get it released?

tests/phpt/handler_should_report_user_errors.phpt
```diff
@@ -12,6 +12,12 @@ trigger_error('abc', E_USER_ERROR);
 
 var_dump('I should not be reached');
 ?>
+--SKIPIF--
+<?php
+if (PHP_VERSION_ID >= 80400) {
+    echo 'SKIP — this test uses methods depreciated in PHP 8.4+';
```
tomlongridge, line 18:
```suggestion
    echo 'SKIP — this test uses methods deprecated in PHP 8.4+';
```

tomlongridge, line 18: I think it would be worth noting this on the PR description – i.e. you can't trigger errors with `E_USER_ERROR`

```suggestion
    echo 'SKIP — this test uses methods deprecated in PHP 8.4+';
```

PR #675, opened 2025-02-26
Redact user fields according to filters

DariaKunoichi:
## Goal
In case user data contains more sensitive information, apply filters to it.
Same process as filtering metadata.

## Changeset
User object wrapped in cleanup function.

## Testing
Manual tests. Added unit tests for user fields filtering.

### DIFF
--- a/src/Report.php
+++ b/src/Report.php
@@ -764,7 +764,7 @@ class Report implements FeatureDataStore
         $event = [
             'app' => $this->config->getAppData(),
             'device' => array_merge(['time' => $this->time], $this->config->getDeviceData()),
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Fixes
+
+* Redact user fields according to filters
```
tomlongridge, line 8:
```suggestion
* Redact user metadata according to redacted keys configuration
```

tomlongridge (approved):

PR #696, opened 2026-07-28
Add support for Guzzle 8

SB-pawanj:
## Goal
Add support for Guzzle 8

## Changeset
Added Guzzle 8 with 7.4 and above versions of PHP

## Testing
All test cases passed

### DIFF
--- a/src/Internal/GuzzleCompat.php
+++ b/src/Internal/GuzzleCompat.php
@@ -37,9 +37,13 @@ final class GuzzleCompat
     /**
      * Get the base URL/URI, which depends on the Guzzle version.
      *
+     * Returns null when using Guzzle 8+ and the provided ClientInterface
+     * implementation does not expose a `getConfig()` method (which was removed
+     * from the interface in Guzzle 8).
+     *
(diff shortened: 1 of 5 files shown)


src/Internal/GuzzleCompat.php
```diff
@@ -37,16 +37,26 @@ public static function getBaseUriOptionName()
...
      * Get the base URL/URI, which depends on the Guzzle version.
      *
+     * Returns null when using Guzzle 8+ and the provided ClientInterface
+     * implementation does not expose a `getConfig()` method (which was removed
+     * from the interface in Guzzle 8).
+     *
      * @param GuzzleHttp\ClientInterface $guzzle
      *
-     * @return mixed
+     * @return mixed|null
```
GrahamCampbell: This change is incorrect. `null` is already contained within `mixed`.

SB-pawanj: Hi @GrahamCampbell, thanks for the review, this has been reverted.

SB-jigneshR (approved):

